PR #6, opened 2021-05-18
Failover

chrisnashdisguise:


Source/RenderStream/Private/RenderStream.cpp
```diff
@@ -270,19 +272,41 @@ bool FRenderStreamModule::PopulateStreamPool()
...
             const FString Channel(description.channel);
-            if (!StreamPool->GetStream(Name))
+            streamInfoArray.Push({ Channel, Name });
+
+            if (TSharedPtr<FFrameStream> Stream = StreamPool->GetStream(Name))
             {
+                // Stream exists in pool but might have changed, update config
+                UE_LOG(LogRenderStream, Log, TEXT("Updating config for existing stream %s at %dx%d"), *Name, Resolution.X, Resolution.Y);
+                check(Resolution == Stream->Resolution());  // Changing stream resolution is not currently supported
+                Stream->Setup(Name, Resolution, Channel, description.clipping, description.handle, description.format);
```
husberg-disguise: Setup calls RSUCHelpers::CreateStreamResources which creates a texture and fence. Texture looks to be refcounted, but fence looks to be a raw pointer.

Source/RenderStream/Private/RenderStream.h
```diff
@@ -38,6 +40,8 @@ class FRenderStreamModule : public IModuleInterface
 
     void EnableStats() const;
 
+    TArray<ARenderStreamEventHandler*> m_eventHandlers;
```
husberg-disguise: Do we need to keep these around or can we look them up before each broadcast? 

If so, may need to use smart pointers to handle lifetimes.

Source/RenderStream/Private/RenderStreamProjectionPolicy.cpp
```diff
@@ -62,12 +62,43 @@ void FRenderStreamProjectionPolicy::StartScene(UWorld* World)
...
+    }
+    else if (APlayerController* NewController = UGameplayStatics::CreatePlayer(World))
+    {
+        PlayerControllerID = UGameplayStatics::GetPlayerControllerID(NewController);
+        Controller = NewController;
+    }
+    else
+        UE_LOG(LogRenderStream, Warning, TEXT("Could not set new view target for capturing."));
+
+    Stream = Module->StreamPool->GetStream(GetViewportId());
```
husberg-disguise, line 83: This seems out of place as the same call is done in ConfigureCapture. 

I'm guessing it's mostly here for the initial PopulateStreamPool call. It would be nice if we could find a better place for it as it should only be called once on startup. If not, I'd just leave it in ConfigureCapture.

husberg-disguise (approved):

Source/RenderStream/Private/RenderStreamProjectionPolicy.cpp
chrisnashdisguise: I'm not sure this matters too much. All it's doing is checking the stream doesn't already exist before populating the stream pool. In this case it probably won't exist yet, but I don't know if there are circumstances in which the stream might have been loaded already.

Source/RenderStream/Private/RenderStream.cpp
chrisnashdisguise: I'm not sure what the implications of this would be. This code isn't really needed for failover any more, as we don't update the config of existing streams, so I'll just remove it if it's likely to cause problems.

Source/RenderStream/Private/RenderStream.h
chrisnashdisguise: It's easiest to get them here while we have access to World and store them. I'll have a look at whether we could store smart pointers instead.

PR #60, opened 2022-08-16
Asset caching performance improvement

tomasgdc:
**DO NOT MERGE as this still needs to be tested by client if it actually helps**

**Issue:**
When using a multiuser editing session with multiple rx's and an editor machine, the constant resaving of the RenderstreamChannelCacheAsset causes the multi user session and Renderstream to crash. This only seems to be an issue when using multiple heavy sublevels as it take a long time to save all the individual cache assets for each sublevel. 

**Profiling:**
During profiling I noticed that **ForceDeleteObjects** function was stalling main thread a lot because it was force replacing references and we don't want to do that.
**AssetCreated** function was taking some amount of time too even though it just broadcasting that new in memory asset is created 

Before changes:
image

After changes:
image

### DIFF
--- a/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
+++ b/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
@@ -129,7 +129,7 @@ void FRenderStreamEditorModule::DeleteCaches(const TArray<FAssetData>& InCachesT
     }
 
     if (Objects.Num() > 0) // Actually stalls for ages even if empty.
-		ObjectTools::DeleteObjectsUnchecked(ObjectsToDelete);
+		ObjectTools::DeleteObjectsUnchecked(Objects);
         //ObjectTools::ForceDeleteObjects(Objects, false);
 }
 
@@ -545,7 +545,8 @@ bool RemoveInvalidCacheEntries()
 
     const auto RemoveCount = ChannelCaches.RemoveAll(IsInvalidCacheAsset);
     if (RemoveCount > 0)
-        ObjectTools::ForceDeleteObjects(ObjectsToDelete, false);
+		ObjectTools::DeleteObjectsUnchecked(ObjectsToDelete);
+        //ObjectTools::ForceDeleteObjects(ObjectsToDelete, false);
 
     return RemoveCount > 0;
 }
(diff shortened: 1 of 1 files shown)


Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -129,7 +129,8 @@ void FRenderStreamEditorModule::DeleteCaches(const TArray<FAssetData>& InCachesT
     }
 
     if (Objects.Num() > 0) // Actually stalls for ages even if empty.
-        ObjectTools::ForceDeleteObjects(Objects, false);
+		ObjectTools::DeleteObjectsUnchecked(Objects);
```
dev-joss: tabs instead of spaces?

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -129,7 +129,8 @@ void FRenderStreamEditorModule::DeleteCaches(const TArray<FAssetData>& InCachesT
...
+        //ObjectTools::ForceDeleteObjects(Objects, false);
```
dev-joss: Remove?

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -486,7 +487,7 @@ URenderStreamChannelCacheAsset* UpdateLevelChannelCache(ULevel* Level)
     // Save the Cache.
     UPackage* Package = Cache->GetPackage();
     Package->MarkPackageDirty();
-    FAssetRegistryModule::AssetCreated(Cache);
+    //FAssetRegistryModule::AssetCreated(Cache);
```
dev-joss: Remove?

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -544,7 +545,8 @@ bool RemoveInvalidCacheEntries()
 
     const auto RemoveCount = ChannelCaches.RemoveAll(IsInvalidCacheAsset);
     if (RemoveCount > 0)
-        ObjectTools::ForceDeleteObjects(ObjectsToDelete, false);
+		ObjectTools::DeleteObjectsUnchecked(ObjectsToDelete);
+        //ObjectTools::ForceDeleteObjects(ObjectsToDelete, false);
```
dev-joss: Remove?

pkevin-disguise:
@tomasgdc this seems quite old and it's targetting 4.27. Can it be closed?

josh-disguise:
@pkevin-disguise Tomas is long since left, would just close.

PR #61, opened 2022-12-06
RSP-247 allow disabling custom event control

josh-disguise:
This patch adds a toggle box in the project plugin settings under 'Disguise Renderstream' called "Detect and control custom events". When this toggle is off, the Renderstream plugin will not export custom events in the schema or attempt to trigger them at runtime. For RSP-247, DSOF-22487


MunWolf:
LGTM, still want to redesign the whole way of doing exposed parameters at some point.

MunWolf (approved):

PR #63, opened 2023-01-23
Fix git info gathering in generate uplugin

JehtBlack:
Fixed some of the ways we gather and use git information to use in the uplugin and builds.
Pushed in an early branch to be merged upwards.


JehtBlack:
Not pull request work, should've just merged locally, apologies

Issue #67, opened 2023-03-17
Missing PostRender/HUD visualization


dandar28:
While working on an integration feature involving visualization of 3d helpers, I noticed that 3d helpers were not displayed when launching project from Disguise, while they are correctly displayed when launching UE project alone.

Here the some debug/dev notes:

> After debugging this I found what caused this:
> It's due to something missing from `URenderStreamViewportClient` implementation with respect to `UGameViewportClient` in the drawing method: the stuff for drawing post render hud is missing from the renderstream implementation.

I already found the involved piece of code that is missing, which could be found in `UGameViewportClient` here : https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Private/GameViewportClient.cpp#L1795-L1815

While it is missing from `URenderStreamViewportClient` and could be placed between these lines: https://github.com/disguise-one/RenderStream-UE/blob/r1.29_UE4.27/Source/RenderStream/Private/RenderStreamViewportClient.cpp#L485-L486

I already found a fix and will be providing a Pull Request asap.

PR #68, opened 2023-03-17
Added missing piece of code for displaying PostRender/HUD stuff in URenderStreamViewportClient

dandar28:
This PR aims to address issue #67 by adding [this piece of code](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Private/GameViewportClient.cpp#L1795-L1815) from `UGameViewportClient` into a compatible moment into the drawing flow of `URenderStreamViewportClient`.


MunWolf:
Looks good to me!

PR #70, opened 2023-05-26
Stutter fix 5.2

ivaylo-disguise:


Source/RenderStream/Private/RSUCHelpers.inl
```diff
@@ -165,9 +165,13 @@ namespace RSUCHelpers
...
         else if (toggle == "D3D12")
         {
+            auto renderPassFence = RHICreateGPUFence(TEXT("RS API Texture Readout"));
+            RHICmdList.WriteGPUFence(renderPassFence);
             {
                 SCOPED_DRAW_EVENTF(RHICmdList, MediaCapture, TEXT("RS Flush"));
-                RHICmdList.ImmediateFlush(EImmediateFlushType::FlushRHIThreadFlushResources);
+                RHICmdList.ImmediateFlush(EImmediateFlushType::FlushRHIThread);
+                while (!renderPassFence->Poll())
+                    FPlatformProcess::Sleep(0.001);
```
josh-disguise: This probably can't be a long-term fix - performance in DX12 mode is dependent on being able to push events into the DX12 queue without synchronising with the CPU threads at all.

The issue is likely caused by us trying to synchronise on a different queue to the one Unreal is rendering on - maybe visible in PIX.

pkevin-disguise:
This was meant as an investigation. The fix is incomplete and should not be merged.

PR #73, opened 2023-06-16
Custom events safety UE 5.0

ivaylo-disguise:
Use an alternative way to invoke custom events to avoid UE CTD on custom event exceptions.

### DIFF
--- a/Source/RenderStream/RenderStream.Build.cs
+++ b/Source/RenderStream/RenderStream.Build.cs
@@ -66,6 +66,8 @@ public class RenderStream : ModuleRules
         //AddEngineThirdPartyPrivateStaticDependencies(Target, "NVAftermath");
         //AddEngineThirdPartyPrivateStaticDependencies(Target, "IntelMetricsDiscovery");	}
 
+        bEnableExceptions = true;
+
         //using (var md5 = MD5.Create())
         //{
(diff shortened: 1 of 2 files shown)


Source/RenderStream/RenderStream.Build.cs
```diff
@@ -66,6 +66,8 @@ public RenderStream(ReadOnlyTargetRules Target) : base(Target)
         //AddEngineThirdPartyPrivateStaticDependencies(Target, "NVAftermath");
         //AddEngineThirdPartyPrivateStaticDependencies(Target, "IntelMetricsDiscovery");	}
 
+        bEnableExceptions = true;
```
dev-joss, line 69: So doesn't throw exceptions without this? UE just handles the error and quits?

ivaylo-disguise: This was added to enable stack unwinding for the plugin and be able to build with the try/catch. Even now, the try/catch doesn't catch anything, but I've left it there just in case for the sake of logging.

The real change is switching from `UFunction::Invoke()` to `AActor::ProcessEvent()`. That seems to change the way UE handles exceptions. Unfortunately, I haven't been able to find more information on the differences between the two, but they both seem to be ways of executing a UFunction.

PR #82, opened 2024-03-18
Fix caches not being generated from all levels

MunWolf:
At one point we added code to collect all maps in a project and generate RenderStreamChannelCache assets if they are missing
(if they already exist we only update the currently loaded ones to minimize save time)

This broke at some point and this is a fix for it.

### DIFF
--- a/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
+++ b/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
@@ -296,7 +296,7 @@ void GenerateParameters(TArray<FRenderStreamExposedParameterEntry>& Parameters, 
             const float Max = HasLimits ? FCString::Atof(*Property->GetMetaData("ClampMax")) : +1;
             CreateField(Parameters.Emplace_GetRef(), Category, Name, "", Name, "", RenderStreamParameterType::Float, Min, Max, 0.001f, v);
         }
-        else if (const FFloatProperty* FloatProperty = CastField<const FFloatProperty>(Property)) 
+        else if (const FFloatProperty* FloatProperty = CastField<const FFloatProperty>(Property))
         {
             const float v = FloatProperty->GetPropertyValue_InContainer(Root);
             UE_LOG(LogRenderStreamEditor, Log, TEXT("Exposed float property: %s is %f"), *Name, v);
@@ -305,7 +305,7 @@ void GenerateParameters(TArray<FRenderStreamExposedParameterEntry>& Parameters, 
             const float Max = HasLimits ? FCString::Atof(*Property->GetMetaData("ClampMax")) : +1;
             CreateField(Parameters.Emplace_GetRef(), Category, Name, "", Name, "", RenderStreamParameterType::Float, Min, Max, 0.001f, v);
         }
-        else if (const FStructProperty* StructProperty = CastField<const FStructProperty>(Property)) 
+        else if (const FStructProperty* StructProperty = CastField<const FStructProperty>(Property))
         {
             const void* StructAddress = StructProperty->ContainerPtrToValuePtr<void>(Root);
             if (StructProperty->Struct == TBaseStructure<FVector>::Get())
@@ -503,7 +503,14 @@ URenderStreamChannelCacheAsset* UpdateLevelChannelCache(ULevel* Level)
     {
         Cache->SubLevels.Empty();
         for (ULevelStreaming* SubLevel : Level->GetWorld()->GetStreamingLevels())
-            Cache->SubLevels.Add(SubLevel->GetWorldAsset()->GetPackage()->GetPathName());
+        {
+            auto WorldAsset = SubLevel->GetWorldAsset();
+            if (WorldAsset == nullptr) {
+                continue;
+            }
+
+            Cache->SubLevels.Add(WorldAsset->GetPackage()->GetPathName());
+        }
     }
 
     // Save the Cache.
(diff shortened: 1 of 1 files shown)


Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -503,7 +503,14 @@ URenderStreamChannelCacheAsset* UpdateLevelChannelCache(ULevel* Level)
     {
         Cache->SubLevels.Empty();
         for (ULevelStreaming* SubLevel : Level->GetWorld()->GetStreamingLevels())
-            Cache->SubLevels.Add(SubLevel->GetWorldAsset()->GetPackage()->GetPathName());
+        {
+            auto WorldAsset = SubLevel->GetWorldAsset();
```
dev-joss:
```
if (auto WorldAsset = SubLevel->GetWorldAsset())
{
    Cache->SubLevels.Add(WorldAsset->GetPackage()->GetPathName());
}
```

might be easier to understand?

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -512,7 +519,8 @@ URenderStreamChannelCacheAsset* UpdateLevelChannelCache(ULevel* Level)
     FAssetRegistryModule::AssetCreated(Cache);
     const FString PackageFileName = FPackageName::LongPackageNameToFilename(CacheFolder + LevelPath, FPackageName::GetAssetPackageExtension());
     FSavePackageArgs args;
-    args.SaveFlags = EObjectFlags::RF_Public | EObjectFlags::RF_Standalone;
+    args.TopLevelFlags = EObjectFlags::RF_Public | EObjectFlags::RF_Standalone;
+    args.bSlowTask = false;
```
dev-joss, line 521: Whats the difference in behavior between this and the previous values?

dev-joss (commented):
Don't understand most of the changes, would need some annotations explaining the changes

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
MunWolf: SaveFlags was the wrong field to set, since both SaveFlags and TopLevelFlags are uint32 instead of the actual enum type this wasn't noticed when the code was migrated to use the FSavePackageArgs structure (which was introduced in 5.0)

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -503,7 +503,12 @@ URenderStreamChannelCacheAsset* UpdateLevelChannelCache(ULevel* Level)
     {
         Cache->SubLevels.Empty();
         for (ULevelStreaming* SubLevel : Level->GetWorld()->GetStreamingLevels())
-            Cache->SubLevels.Add(SubLevel->GetWorldAsset()->GetPackage()->GetPathName());
+        {
+            if (auto WorldAsset = SubLevel->GetWorldAsset())
+            {
+                Cache->SubLevels.Add(WorldAsset->GetPackage()->GetPathName());
+            }
+        }
```
MunWolf, line 511: I noticed an issue where some sublevels did not have an asasociated world asset
(seemed to be some auto generated world streaming thing)
Since they don't have a world they are not saved as an actual level so we don't care about it.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
MunWolf: Slow task was also set correctly in 4.27 but the default value was wrong in the structure.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -589,15 +592,17 @@ void UpdateChannelCache()
 
     // Loop over all levels and make sure caches exist for them.
     TArray<FAssetData> LevelAssets;
-    const auto LevelLibrary = UObjectLibrary::CreateLibrary(ULevel::StaticClass(), false, true);
+    const auto LevelLibrary = UObjectLibrary::CreateLibrary(UWorld::StaticClass(), false, true);
     LevelLibrary->LoadAssetDataFromPath(ContentFolder);
```
MunWolf, line 596: This is the important part, apparently at some point they changed it so you can no longer load levels directly but you have to load worlds instead (which are technically a collection of levels but the persistant level which we care about is always the first one)

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -830,7 +835,9 @@ void FRenderStreamEditorModule::OnObjectPostEditChange(UObject* Object, FPropert
     {
         // we only care if default objects have been changed eg. project settings objects like UGameMapSettings
         // add include/exclude filters here if required
-        DirtyAssetMetadata = true;
+        if (Object->IsA<UGameMapsSettings>() || Object->IsA<URenderStreamSettings>()) {
+            DirtyAssetMetadata = true;
+        }
```
MunWolf, line 840: This is to prevent it from regenerating the schema every time any project setting is changed in unreal and limiting it to only relevant ones

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
pkevin-disguise: Would that work on a 4.27 project?

I know the plugin says 5.1, but we know of setups running 4.27 project on 5.1 plugins.

MunWolf: Yes, the main difference between the 4.27 version of this change and 5.x is the FSavePackageArg above doesn't exist.
But this one will work like this (Jordan confirmed the current released version of 4.27 does not work)

pkevin-disguise (approved):
lgtm

dev-joss (approved):
No red flags from me, not an unreal expert however

PR #85, opened 2024-05-03
Plugin MultiUserClient Fails To Load When Launching A Project With The RenderStream Plugin

josharms-disguise:
A plugin that MultiUserClient requires does not get enabled by default so this change forces that plugin (ConcertSyncClient) to be enabled in the plugin template so everything can load properly.

### DIFF
--- a/uplugin_template.json
+++ b/uplugin_template.json
@@ -49,6 +49,10 @@
 		{
 			"Name": "nDisplay",
 			"Enabled": true
+		},
+		{
+			"Name": "ConcertSyncClient",
+			"Enabled": true
 		}
 	]
 }


twhittock-disguise:
we don't depend on multi user client, right? this should only be enabled when the multi user client plugin is enabled... is there nowhere we can do that? I'd prefer not to bloat our dependencies for something optional.

josharms-disguise:
> we don't depend on multi user client, right? this should only be enabled when the multi user client plugin is enabled... is there nowhere we can do that? I'd prefer not to bloat our dependencies for something optional.

It's required for nDisplay but in 5.4 it doesn't get enabled automatically for some reason

pkevin-disguise (approved):
It's a little bit strange that we have to manually load a plugin that the multi user plugin depend on and not manually load the multi user plugin directly, but I guess the multi user plugin isn't enabling its dependencies correctly.

LGTM

josharms-disguise:
> It's a little bit strange that we have to manually load a plugin that the multi user plugin depend on and not manually load the multi user plugin directly, but I guess the multi user plugin isn't enabling its dependencies correctly.
> 
> LGTM

I have been told they are planning to fix this in 5.5

Jordan-Goulet-Disguise-One:
**QA passed.**

Jordan-Goulet-Disguise-One (approved):
QA passed.

PR #90, opened 2024-09-11
Adds nullptr checking when caching levels

ivaylo-disguise:
Assets are loaded and cast to UWorld. Under certain situations, such as copy-pasting world assets manually, this procedure may return a nullptr. Without nullptr checking, an access violation crashes Unreal Engine.

Here we catch nullptr and log an appropriate error message for the user.

RSP-348


josharms-disguise (approved):
lgtm

Jordan-Goulet-Disguise-One (approved):
**QA Passed**

MunWolf:
Looks like there is an issue with spacing (tabs vs spaces)

ivaylo-disguise:
> Looks like there is an issue with spacing (tabs vs spaces)

true, I'll check what happened there.

PR #91, opened 2024-11-13
Updated plugin to be compatible with UE 5.5

josharms-disguise:
[RSP-351](https://d3technologies.atlassian.net/browse/RSP-351)

**Technical Description:**
The RenderStream UE plugin currently only supports up to UE 5.4, this change allows the plugin to be compatible with UE 5.5. The FRHIBatchedShaderParameters struct is no longer allowed to be default constructed. There is also an issue with settings not being able to be saved to the DefaultEngine.ini file. This is because of some unknown issue with using SaveConfig() on these settings.

**Resolution Details:**
We can instantiate the FRHIBatchedShaderParameters struct using the command list that is already being passed into the function where the issue is. 

The workaround for saving settings is to not use SaveConfig(), currently we use a customization class to extend the SceneSelection setting so that when it is changed we can have a callback to generate new metadata. This however also requires the use of SaveConfig() in this callback since it won't be saved otherwise. The solution is just to exclude this customization code for now since the plugin already is checking separately for changes in the settings and will generate the metadata on shutdown. See FRenderStreamEditorModule::OnObjectPostEditChange and FRenderStreamEditorModule::OnShutdownPostPackagesSaved(). 

**Test Guidance:**
UE 5.5 assets should work with RenderStream. The dx11 issue was fixed with UE version 5.5.1.

[RSP-351]: https://d3technologies.atlassian.net/browse/RSP-351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStream/Private/RSUCHelpers.inl
+++ b/Source/RenderStream/Private/RSUCHelpers.inl
@@ -79,7 +79,7 @@ namespace {
         }
 
         TUniformBufferRef<RSResizeCopyUB> Data = TUniformBufferRef<RSResizeCopyUB>::CreateUniformBufferImmediate(UB, UniformBuffer_SingleFrame);
-        FRHIBatchedShaderParameters Params;
+        FRHIBatchedShaderParameters Params = CommandList.GetScratchShaderParameters();
         SetUniformBufferParameter(Params, GetUniformBufferParameter<RSResizeCopyUB>(), Data);
         CommandList.SetBatchedShaderParameters(CommandList.GetBoundPixelShader(), Params);
     }
--- a/Source/RenderStream/Public/RenderStreamLink.h
+++ b/Source/RenderStream/Public/RenderStreamLink.h
@@ -28,7 +28,7 @@ typedef uint64_t VkDeviceSize;
 typedef struct VkSemaphore_T* VkSemaphore;
 
 #define RS_PLUGIN_NAME "RenderStream-UE"
-#define RS_PLUGIN_VERSION "RS2.0-UE5.5-v1"
+#define RS_PLUGIN_VERSION "RS2.0-UE5.5-v0"
 
 class RenderStreamLink
 {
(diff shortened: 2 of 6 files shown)


Source/RenderStream/Private/RSUCHelpers.inl
```diff
@@ -79,7 +79,7 @@ namespace {
         }
 
         TUniformBufferRef<RSResizeCopyUB> Data = TUniformBufferRef<RSResizeCopyUB>::CreateUniformBufferImmediate(UB, UniformBuffer_SingleFrame);
-        FRHIBatchedShaderParameters Params;
+        FRHIBatchedShaderParameters Params = CommandList.GetScratchShaderParameters();
```
pkevin-disguise, line 82: that's all it took in the end? 🤯

pkevin-disguise (approved):
lgtm

pkevin-disguise (changes requested):
I don't think we should merge this in 5.4 though

josharms-disguise:
> I don't think we should merge this in 5.4 though

I was just going to close the PR when it got approved, there is no release branch so I'm not sure what the normal protocol is

pkevin-disguise (commented):
lgtm then

pkevin-disguise (approved):
lgtm

Source/RenderStream/Public/RenderStreamLink.h
```diff
@@ -28,7 +28,7 @@ typedef uint64_t VkDeviceSize;
 typedef struct VkSemaphore_T* VkSemaphore;
 
 #define RS_PLUGIN_NAME "RenderStream-UE"
-#define RS_PLUGIN_VERSION "RS2.0-UE5.4-v1"
+#define RS_PLUGIN_VERSION "RS2.0-UE5.5-v1"
```
egorovaanastasiia, line 31: Quick q: should it be v1 or v0?

josharms-disguise: Looking back at old versions it looks like it should be v0, I will change it

egorovaanastasiia (approved):
lgtm

ivaylo-disguise (approved):

pkevin-disguise:
This plugin has been released.

PR #92, opened 2024-12-11
Checkout cache files when locked by version control

ivaylo-disguise:
If cache files are marked as read-only by version control, trying to write into them crashed UE editor.
Open the checkout window instead for all of the cached packages to try and make these cache files writeable. On failure, skip saving the RS schema and log error.


josharms-disguise (approved):
lgtm

ivaylo-disguise:
Ticket will be merged into UE5.2 and cascaded upwards from there.
See https://github.com/disguise-one/RenderStream-UE/pull/94

PR #95, opened 2024-12-13
Exposed parameters don't work

egorovaanastasiia:
[RSP-356](https://d3technologies.atlassian.net/browse/RSP-356)

**Technical Description:**
Exposed parameters didn't work when adding another exposed parameter to a sublevel while UE project is on Maps Scene Selector mode. The main problem was with failing  `ValidateParameters` function that caused not calling `ApplyParameters`. I noticed that `ValidateParameters` was failing because JSON provided more parameters than we expected. 

Another issue was with text parameters values worked only for the 1st text param, but not for 2nd+ ones. That happened because we didn't have an internal count and we were always reading the 0 value.

**Resolution Details:**
The main issue was with the way we generated the JSON file for different Scene Selectors. I changed it, so now it happens according to the following table:
| Scene Selector | Persistent Level | Sublevel |
| ------------- | ------------- | ------------- |
| None  | Parameters from all levels  | X  |
| Maps  | Persistent level's parameters only  | Sublevel's parameters only |
| Streaming Level | Persistent level's parameters only  | Persistent level's + Sublevel's parameters |

Source: https://help.disguise.one/workflows/renderstream/unreal-engine/scene-levels (which is imo a bit confusing and should be updated as well)

For the text parameters issue, I added a count similar to the ones we have for floats and images.

[RSP-356]: https://d3technologies.atlassian.net/browse/RSP-356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
+++ b/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
@@ -451,18 +451,19 @@ void RenderStreamSceneSelector::ApplyParameters(uint32_t sceneId, const TArray<A
     // These are updated by ApplyParameters to allow each actor to operate on the next set of data.
     const RenderStreamLink::RemoteParameter* paramsPtr = params.parameters;
     const float* floatValuesPtr = floatValues.data();
+    size_t textValues = 0;
     const RenderStreamLink::ImageFrameData* imageValuesPtr = imageValues.data();
     for (AActor* actor : Actors)
     {
         if (!actor)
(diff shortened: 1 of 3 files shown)


MunWolf:
LGTM 👍

ivaylo-disguise (approved):

Jordan-Goulet-Disguise-One (approved):
**QA PASSED**

PR #97, opened 2025-01-09
Post Processing Settings Being Applied To The Wrong Camera

josharms-disguise:
Technical Description

In a scene with 2 cameras and different post processing setting, when using both of those cameras within the same layer the settings seem to be swapped between the cameras. The workaround is to have each camera as a single layer but this hurts performance. This solution fixes the issue by making sure the settings are being applied to the current camera before everything is submitted for rendering.


ivaylo-disguise (approved):

pkevin-disguise (approved):
lgtm

Jordan-Goulet-Disguise-One (approved):
**QA PASSED**

PR #98, opened 2025-01-22
RSP-345: Camera focus does not work in UE 5.4

ivaylo-disguise:
## Technical Description:
UE 5.4 implements new ICVFX features which mean that the nDisplay viewports and projections code has changed in regards to post-processing.
An example can be seen in the nDisplay camera projection code where camera post-processing parameters are set on every viewport in a new virtual function called `UpdatePostProcessSettings`.
The RenderStream projection policy does not override that function so it must be using default post processing settings.

## Resolution Details:
I have implemented the `UpdatePostProcessSettings` function similarly to how it is implemented in `FDisplayClusterProjectionCameraPolicy`. In the case of RenderStream, I have enabled camera post-processing by default.


pkevin-disguise (approved):
lgtm

josharms-disguise (approved):
lgtm

Jordan-Goulet-Disguise-One (approved):
**QA PASSED**

PR #99, opened 2025-04-10
Launch Unreal Engine Exes Through RenderStream

josharms-disguise:
Technical Description

The main issue with getting Unreal exes to launch through RenderStream is getting the metadata .json over to where the newly created exe is. This is done by creating a separate button in the UI that the plugin controls. This button will start the packaging process for Windows platforms and then will copy the metadata over once finished. The second issue is when loading the schema the plugin only checks in the project directory, to fix this it now checks if it is running the packaged version or not and sends the correct asset path. 

~~One outstanding issue is getting the Unreal project to package correctly. For some reason when running the packaged project it complains that it can't find the nDisplay plugin. So far I have not found a way to guarantee that this plugin is packaged with the exe.~~

### DIFF
--- a/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
+++ b/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
@@ -5,7 +5,6 @@
 #include <malloc.h>
 #include "RenderStream.h"
 #include "RenderStreamHelper.h"
-#include "RSUCHelpers.inl"
 #include "RenderStreamSettings.h"
 #include "Engine/LevelStreaming.h"
 #include "Engine/LevelScriptActor.h"
@@ -13,6 +12,16 @@
 #include "TextureResource.h"
 
 #include "ProfilingDebugging/RealtimeGPUProfiler.h"
+
+#include "Engine/Level.h"
+#include "Engine/World.h"
+#include "UObject/TextProperty.h"
+
+#include "VulkanRHIPrivate.h"
+#include "VulkanRHIBridge.h"
+#include "VulkanResources.h"
+
+#include "HardwareInfo.h"
 
 RenderStreamSceneSelector::~RenderStreamSceneSelector() = default;
 
--- a/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
+++ b/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
@@ -101,23 +101,14 @@ const RenderStreamLink::Schema& RenderStreamSceneSelector::Schema() const
         return m_defaultSchema.schema;
 }
 
-bool RenderStreamSceneSelector::IsExe() const
-{
-    FString executableName = FPlatformProcess::ExecutableName();
-    return !executableName.Contains("UnrealEditor");
-}
-
 void RenderStreamSceneSelector::LoadSchemas(const UWorld& World)
 {
-    std::string AssetPath;
-    if(IsExe())
-    {
-        AssetPath = TCHAR_TO_UTF8(FPlatformProcess::ExecutablePath());
-    }
-    else
-    {
-        AssetPath = TCHAR_TO_UTF8(*FPaths::GetProjectFilePath());
-    }
+    #if WITH_EDITOR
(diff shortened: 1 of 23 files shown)


Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -859,6 +865,106 @@ void FRenderStreamEditorModule::GenerateAssetMetadata()
...
+        -platform=Win64 -installed -stage -archive -package -build -pak -iostore -compressed -prereqs \
+        -archivedirectory=\"%s\" -clientconfig=Development -nocompile -nocompileuat"),
+        *projectPath,
+        *enginePath,
+        *outputFolder);
+
+    FString errorOut;
+    int32 ReturnCode = 0;
+    UE_LOG(LogTemp, Log, TEXT("Packaging started..."));
+    bool bSuccess = FPlatformProcess::ExecProcess(*uatPath, *arguments, &ReturnCode, &errorOut, nullptr);
```
josharms-disguise, line 933: Currently this is blocking, can be changed if we think it will be too annoying

josharms-disguise:
I think I might have found the reason that it isn't being packaged properly. If I include an nDisplay config it seems to work. Also as far as I can tell an nDisplay config cannot be retroactively applied to packaged exes and needs to be made before the build. That means the current workflow would be running the desired asset through RenderStream -> nDisplay config is generated -> package project -> run exe through RenderStream. I'll have to look at possible workarounds.

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -101,10 +101,23 @@ const RenderStreamLink::Schema& RenderStreamSceneSelector::Schema() const
         return m_defaultSchema.schema;
 }
 
+bool RenderStreamSceneSelector::IsExe() const
+{
+    FString executableName = FPlatformProcess::ExecutableName();
+    return !executableName.Contains("UnrealEditor");
```
ivaylo-disguise: Would using `#if WITH_EDITOR` work? If it works, it could even replace the whole function and just use the macro in LoadSchemas().

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
ivaylo-disguise: Is there any visual feedback to the user that packaging is in progress?

ivaylo-disguise (approved):
Pretty cool

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
josharms-disguise: Nice, that seems to work well

josharms-disguise:
@ivaylo-disguise I can't respond directly to your other comment for some reason. Currently the command prompt will pop up and spit out what's going on in the build process. Not sure if it's possible to do something fancier like a progress bar or something

egorovaanastasiia (approved):
lgtm

egorovaanastasiia:
nDisplay plugin issue is not critical for now, right?

josharms-disguise:
> nDisplay plugin issue is not critical for now, right?

As in we don't need to worry about finding a workaround right now ? I'm trying to find a set of reproducible steps right now that will always package a fresh project properly. I'm not sure if we can ever get this process to be fully automatic but we should try reducing it to as least steps as possible

josharms-disguise:
> I think I might have found the reason that it isn't being packaged properly. If I include an nDisplay config it seems to work. Also as far as I can tell an nDisplay config cannot be retroactively applied to packaged exes and needs to be made before the build. That means the current workflow would be running the desired asset through RenderStream -> nDisplay config is generated -> package project -> run exe through RenderStream. I'll have to look at possible workarounds.

Disregard this, adding LiveLinkOverNDisplay as a dependency resolves this issue, nothing to do with having an nDisplay config or not

Jordan-Goulet-Disguise-One (approved):
**QA PASSED**

PR #104, opened 2025-06-05
UE DX11 Assets Do Not Work With Unreal 5.5 and Texture Parameters

josharms-disguise:
(RSP-373)[https://d3technologies.atlassian.net/browse/RSP-373]

When running RenderStream using UE 5.5 and the dx11 backend the workload immediately crashes. This is because dx11 complains that the context is being accessed from 2 different threads at the same time. The fix is to call an immediate flush before grabbing the texture parameter. This executes all the queued render commands before executing ours.


pkevin-disguise (approved):
we probably should pay attention to performance when testing this fix.

egorovaanastasiia (approved):
lgtm

Jordan-Goulet-Disguise-One (approved):
**QA PASSED**

PR #105, opened 2025-06-16
UE crashes when starting a workload with custom events and a sub level and scene selection set to None

josharms-disguise:
[RSP-361](https://d3technologies.atlassian.net/browse/RSP-361?atlOrigin=eyJpIjoiZGIzYTA3Mjg4NWVhNDU5ZDhkYTRjMTMwZTVlZGEzN2MiLCJwIjoiaiJ9)

**Technical Description:**

When the scene selection is set to None then the sublevels are not loaded. However in the plugin, the sublevels of a persistent level are added to the cache regardless of what the mode is set to. This causes any customs events present in the sublevel to be saved to the schema which then causes UE to crash on startup because it is trying to find a custom event for a level that was never loaded. The fix for this is just to check the scene selection mode when updating the level cache.

[RSP-361]: https://d3technologies.atlassian.net/browse/RSP-361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
+++ b/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
@@ -546,15 +546,22 @@ URenderStreamChannelCacheAsset* UpdateLevelChannelCache(ULevel* Level)
     Cache->ExposedParams.Empty();
     GenerateParameters(Cache->ExposedParams, Level->GetLevelScriptActor());
 
+    const URenderStreamSettings* settings = GetDefault<URenderStreamSettings>();
+
     // We can only know the sublevels of the persistent level.
     if (Level->IsPersistentLevel())
     {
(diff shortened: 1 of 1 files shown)


egorovaanastasiia (approved):
lgtm.

Should we make a fix for earlier versions of UE though? Seems like some clients are still using UE5.3 projects

josharms-disguise:
> lgtm.
> 
> Should we make a fix for earlier versions of UE though? Seems like some clients are still using UE5.3 projects

Ya we can, I was also going to add this fix to the later versions as well

ivaylo-disguise (approved):
lgtm

Jordan-Goulet-Disguise-One (approved):
**QA PASSED**

PR #109, opened 2025-06-18
Custom events do not work correctly on sub levels when streaming levels in RenderStream

josharms-disguise:
[RSP-360](https://d3technologies.atlassian.net/browse/RSP-360?atlOrigin=eyJpIjoiZWUwY2U2Yzk2NWUyNDAzYTgxMTA2MjBhOTEyNWJkY2MiLCJwIjoiaiJ9)

**Technical Description**

When using the StreamingLevels SceneSelection mode and while on the Sublevel scene, the persistent level’s parameters will also be included. There is a loop in RenderStreamSceneSelector::ApplyParameters that iterates over all the LevelActors and calls ApplyParameters (a different one) on each level. The way this loop handles the float parameters that it gets them from d3 in one vector (parameters from all levels are included in this) then uses a counter (iFloat) to index into that float array to get the proper float. This counter is reset at the start of each ApplyParameter call so that each level will start will iFloat = 0. At the end of each call the float pointer array is incremented by the amount of iFloat so that it will point to the correct float when the next level (sublevel in this case) has ApplyParameters called for it. This would work except that when this function checks if the event should be invoked it checks the previous floats vs the new floats it just got from d3. If the counter has increased then the event should be invoked. However the array that is indexed to check the old values does not behave the same way as we get the new value. The old value array is passed with all the previous float values, so when using this reset iFloat counter it will take the previous level’s old custom event counter and compare it to the current level’s new event counter. This causes the sublevel’s event to never be invoked. The fix makes the counter persistent while looping over the LevelActors so that it won’t be reset in-between calls of ApplyParameters allowing the retrieval of the correct float value.

[RSP-360]: https://d3technologies.atlassian.net/browse/RSP-360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
+++ b/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
@@ -499,20 +499,20 @@ void RenderStreamSceneSelector::ApplyParameters(uint32_t sceneId, const TArray<A
 
     // These are updated by ApplyParameters to allow each actor to operate on the next set of data.
     const RenderStreamLink::RemoteParameter* paramsPtr = params.parameters;
-    const float* floatValuesPtr = floatValues.data();
     size_t textValues = 0;
     const RenderStreamLink::ImageFrameData* imageValuesPtr = imageValues.data();
+    size_t iFloat = 0;
     for (AActor* actor : Actors)
     {
         if (!actor)
             continue; // it's convenient at the higher level to pass nulls if there's a pattern which can miss pieces
-        ApplyParameters(actor, params.hash, &paramsPtr, params.nParameters, &floatValuesPtr, floatValues.size(), &imageValuesPtr, imageValues.size(), textValues);
+        ApplyParameters(actor, params.hash, &paramsPtr, params.nParameters, floatValues, iFloat, &imageValuesPtr, imageValues.size(), textValues);
     }
 
     m_floatValuesLast = floatValues; // event parameters need to lookup previous values
 }
 
-void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash, const RenderStreamLink::RemoteParameter** ppParams, const size_t nParams, const float** ppFloatValues, const size_t nFloatVals, const RenderStreamLink::ImageFrameData** ppImageValues, const size_t nImageVals, size_t& textValues)
+void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash, const RenderStreamLink::RemoteParameter** ppParams, const size_t nParams, const std::vector<float>& floatValues, size_t& iFloat, const RenderStreamLink::ImageFrameData** ppImageValues, const size_t nImageVals, size_t& textValues)
 {
     auto toggle = FHardwareInfo::GetHardwareInfo(NAME_RHI);
     struct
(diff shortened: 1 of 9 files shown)


egorovaanastasiia (approved):
lgtm

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -499,20 +499,20 @@ void RenderStreamSceneSelector::ApplyParameters(uint32_t sceneId, const TArray<A
...
             continue; // it's convenient at the higher level to pass nulls if there's a pattern which can miss pieces
-        ApplyParameters(actor, params.hash, &paramsPtr, params.nParameters, &floatValuesPtr, floatValues.size(), &imageValuesPtr, imageValues.size(), textValues);
+        ApplyParameters(actor, params.hash, &paramsPtr, params.nParameters, floatValues, iFloat, &imageValuesPtr, imageValues.size(), textValues);
     }
 
     m_floatValuesLast = floatValues; // event parameters need to lookup previous values
 }
 
-void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash, const RenderStreamLink::RemoteParameter** ppParams, const size_t nParams, const float** ppFloatValues, const size_t nFloatVals, const RenderStreamLink::ImageFrameData** ppImageValues, const size_t nImageVals, size_t& textValues)
+void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash, const RenderStreamLink::RemoteParameter** ppParams, const size_t nParams, const std::vector<float>& floatValues, size_t& iFloat, const RenderStreamLink::ImageFrameData** ppImageValues, const size_t nImageVals, size_t& textValues)
```
pkevin-disguise: Is there something I am missing regarding the iFloat parameter? size_t and size_t& is the same on 64bit architecture. I also don't see it being set in the function and it isn't a const ref.

In theory, copying the value would be safer but in this case I don't think it changes anything since it's all in scope.

pkevin-disguise (commented):
lgtm

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
josharms-disguise: I wanted as a reference so it could be modified in ApplyParameters, it does get incremented once when it tries to invoke the event and another time when it sets the exposed property. So In the scenario that I had (Persistent Level w/ Custom Event and SubLevel w/ Custom Event), with the old way iFloat would be 0 again when it checks to see if the SubLevel event can be invoked. But if iFloat is 0 then it's looking at the first element of m_floatValuesLast which would correspond to the Persistent Level's custom event counter.

pkevin-disguise: ah right, I wasn't seeing it as a change because the incrementation was old code. Makes sense now, thanks!

pkevin-disguise (approved):
lgtm

ivaylo-disguise (commented):
lgtm

Jordan-Goulet-Disguise-One (approved):
**QA PASSED**

pkevin-disguise:
Fix came from 5.3 merge.

PR #119, opened 2025-11-03
HDR Colour Transform for Texture Parameters

egorovaanastasiia:
[DSOF-31260](https://d3technologies.atlassian.net/browse/DSOF-31260)

**Bug Description:**
We didn't support texture parameter colour transform, including HDR colours. Changes for this ticket implement it. 

**Resolution Details:**
This ticket consists of 2 parts: UE plugin changes (this PR) and d3 changes ([that PR](https://github.com/disguise-one/d3/pull/5174).)

The current flow of the texture param colour transform looks like:

1. When sending texture param from d3, create a new output colour space, that would match the "Input transform" of the UE project (d3 related changes in a different PR.)
2. When receiving a frame in UE plugin, we use a second RenderTarget-helper of a format that supports texture stitching, while setting the original Texture's to a format that supports HDR colours (`EPixelFormat::PF_FloatRGBA`).
3. ~Copy the original UE provided Texture into the RenderTarget helper.~(Don't need this step actually.)
4. Copy the texture parameter into the RenderTarget helper.
5. Apply the colour transform from the helper to the final UE Texture.

Original flow with no colour transform remained unchanged, I only refactored some code.

Note:
During final testing I had a problem with workload crashing after running for a minute. I was able to resolve it after adding the new RenderTarget helper object to the Root. This allows to protect the object form the Garbage Collector, which is crucial since it's also used in the Render thread.

**Notes For QA:**
The ticket should be reproducible with the files and steps to reproduce that Effy mentioned. The only additional step is to enable `Support inverse view transform` setting under `Plugins - OpenColorIO` in the UE project.

The change includes 2 parts: d3 and UE plugin changes. The fix works only with the correct version of both.

Along with the correct colour transformation, please check if the RS workload is stable. 

[DSOF-31260]: https://d3technologies.atlassian.net/browse/DSOF-31260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
+++ b/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
@@ -20,7 +20,13 @@
 #include <OpenColorIOColorTransform.h>
 #include <OpenColorIORendering.h>
 
-RenderStreamSceneSelector::~RenderStreamSceneSelector() = default;
+RenderStreamSceneSelector::~RenderStreamSceneSelector()
+{
+    if (m_textureColourTransform && m_textureColourTransform->IsRooted())
+    {
+        m_textureColourTransform->RemoveFromRoot();
+    }
+}
 
 void RenderStreamSceneSelector::GetAllLevels(TArray<AActor*>& Actors, ULevel * Level) const
 {
@@ -201,6 +207,27 @@ bool RenderStreamSceneSelector::ValidateParameters(const RenderStreamLink::Remot
             Actors.Num(), offset, sceneParameters.nParameters);
         return false;
     }
+
+    const URenderStreamSettings* settings = GetDefault<URenderStreamSettings>();
+    if (settings->OCIOConfig.bIsEnabled && settings->OCIOConfig.ColorConfiguration.ConfigurationSource)
+    {
+        isColourConfigurationEnabled = true;
+
+        m_colourConversionSettings.ConfigurationSource = settings->OCIOConfig.ColorConfiguration.ConfigurationSource;
+
+        if (settings->OCIOConfig.ColorConfiguration.DestinationColorSpace.ColorSpaceIndex != INDEX_NONE)
+        {
+            m_colourConversionSettings.SourceColorSpace = settings->OCIOConfig.ColorConfiguration.DestinationColorSpace;
+            m_colourConversionSettings.DestinationColorSpace = settings->OCIOConfig.ColorConfiguration.SourceColorSpace;
+        }
+        else
+        {
+            m_colourConversionSettings.SourceColorSpace = settings->OCIOConfig.ColorConfiguration.SourceColorSpace;
+            m_colourConversionSettings.DestinationDisplayView = settings->OCIOConfig.ColorConfiguration.DestinationDisplayView;
+            m_colourConversionSettings.DisplayViewDirection = EOpenColorIOViewTransformDirection::Inverse;
+        }
+    }
+
 
     return true;
 }
@@ -520,6 +547,74 @@ void RenderStreamSceneSelector::ApplyParameters(uint32_t sceneId, const TArray<A
     m_floatValuesLast = floatValues; // event parameters need to lookup previous values
 }
 
+void RenderStreamSceneSelector::GetTextureParameter(const FString& toggle, const RenderStreamLink::ImageFrameData& frameData, size_t iImage, UTextureRenderTarget2D* Texture)
+{
+    ENQUEUE_RENDER_COMMAND(GetTex)(
+        [Texture, toggle, frameData, iImage](FRHICommandListImmediate& RHICmdList)
+        {
+            SCOPED_DRAW_EVENTF(RHICmdList, MediaCapture, TEXT("RS Tex Param Block %d"), iImage);
+            const auto rtResource = Texture->GetRenderTargetResource();
+            if (!rtResource)
+            {
+                return;
+            }
+            void* resource = rtResource->TextureRHI->GetNativeResource();
+
+            RenderStreamLink::SenderFrame data = {};
+            if (toggle == "D3D11")
+            {
+                data.type = RenderStreamLink::SenderFrameType::RS_FRAMETYPE_DX11_TEXTURE;
+                data.dx11.resource = static_cast<ID3D11Resource*>(resource);
+                auto err = RenderStreamLink::instance().rs_getFrameImage2(frameData.imageId, &data);
+            }
+            else if (toggle == "D3D12")
+            {
+                {
+                    SCOPED_DRAW_EVENTF(RHICmdList, MediaCapture, TEXT("RS Tex Param Flush"));
+                    RHICmdList.ImmediateFlush(EImmediateFlushType::FlushRHIThreadFlushResources);
+                }
+
+                data.type = RenderStreamLink::SenderFrameType::RS_FRAMETYPE_DX12_TEXTURE;
+                data.dx12.resource = static_cast<ID3D12Resource*>(resource);
+
+                SCOPED_DRAW_EVENTF(RHICmdList, MediaCapture, TEXT("RS getFrameImage2 %d"), iImage);
+                if (RenderStreamLink::instance().rs_getFrameImage2(frameData.imageId, &data) != RenderStreamLink::RS_ERROR_SUCCESS)
+                {
+
+                }
+            }
+            else if (toggle == "Vulkan")
+            {
+                {
+                    SCOPED_DRAW_EVENTF(RHICmdList, MediaCapture, TEXT("RS Tex Param Flush"));
+                    RHICmdList.ImmediateFlush(EImmediateFlushType::FlushRHIThreadFlushResources);
+                }
+
+                FVulkanTexture* VulkanTexture = ResourceCast(rtResource->TextureRHI->GetTexture2D());
+                auto point2 = VulkanTexture->GetSizeXY();
+
+                data.type = RenderStreamLink::SenderFrameType::RS_FRAMETYPE_VULKAN_TEXTURE;
+                data.vk.memory = VulkanTexture->GetAllocationHandle();
+                data.vk.size = VulkanTexture->GetAllocationOffset() + VulkanTexture->GetMemorySize();
+                data.vk.format = frameData.format;
+                data.vk.width = uint32_t(point2.X);
+                data.vk.height = uint32_t(point2.Y);
+                // TODO: semaphores
+
+                SCOPED_DRAW_EVENTF(RHICmdList, MediaCapture, TEXT("RS getFrameImage2 %d"), iImage);
+                if (RenderStreamLink::instance().rs_getFrameImage2(frameData.imageId, &data) != RenderStreamLink::RS_ERROR_SUCCESS)
+                {
+
+                }
+            }
+            else
+            {
+                UE_LOG(LogRenderStream, Error, TEXT("RenderStream tried to send frame with unsupported RHI backend."));
+                return;
+            }
+        });
+}
+
 void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash, const RenderStreamLink::RemoteParameter** ppParams, const size_t nParams, const std::vector<float>& floatValues, size_t& iFloat, const RenderStreamLink::ImageFrameData** ppImageValues, const size_t nImageVals, size_t& textValues)
 {
     auto toggle = FHardwareInfo::GetHardwareInfo(NAME_RHI);
@@ -667,6 +762,12 @@ void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash,
             UObject* o = ObjectProperty->GetObjectPropertyValue(ObjectAddress);
             if (UTextureRenderTarget2D* Texture = Cast<UTextureRenderTarget2D>(o))
             {
+                if (iImage >= nImageVals)
+                {
+                    UE_LOG(LogRenderStream, Verbose, TEXT("Attempt to read a image value from disguise that is out of range. Does the metadata need to be regenerated?"));
+                    continue;
+                }
+
                 // Treat all incoming textures are in linear colour space to
                 // ensure consistent behaviour.
                 // Without this, UE applies gamma 2.2 to textures with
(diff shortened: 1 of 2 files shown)


Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -653,94 +754,115 @@ void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash,
...
-                        if (RenderStreamLink::instance().rs_getFrameImage2(frameData.imageId, &data) != RenderStreamLink::RS_ERROR_SUCCESS)
-                        {
+                    if (!m_textureColourTransform->bGPUSharedFlag || m_textureColourTransform->GetFormat() != formatMap[frameData.format].ue)
+                    {
+                        m_textureColourTransform->bGPUSharedFlag = true;
+                        m_textureColourTransform->InitCustomFormat(frameData.width, frameData.height, formatMap[frameData.format].ue, false);
+                    }
+                    else
+                    {
+                        Texture->ResizeTarget(frameData.width, frameData.height);
```
josharms-disguise: Did you mean to resize the textureColourTransform here ?

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -653,94 +754,115 @@ void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash,
...
+                            Root,
+                            Texture->SizeX,
+                            Texture->SizeY,
+                            Texture->RenderTargetFormat
+                        );
+
+                        m_textureColourTransform->SRGB = 0;
+                        m_textureColourTransform->bForceLinearGamma = true;
+
+                        m_textureColourTransform->AddToRoot();
```
josharms-disguise: Do we ever need to call RemoveFromRoot at some point ?

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -13,8 +13,20 @@
...
+#include "Materials/MaterialInstanceDynamic.h"
+#include "OpenColorIOBlueprintLibrary.h"
+#include <OpenColorIOColorTransform.h>
+#include <OpenColorIORendering.h>
+
+RenderStreamSceneSelector::~RenderStreamSceneSelector()
+{
+    if (m_textureColourTransform && m_textureColourTransform->IsRooted())
+    {
+        m_textureColourTransform->RemoveFromRoot();
```
DevTGHa: Do we want to release `m_textureColourTransform`?
For example:
```cpp
    if (m_textureColourTransform)
    {
        if(m_textureColourTransform->IsRooted())
            m_textureColourTransform->RemoveFromRoot();

        UKismetRenderingLibrary::ReleaseRenderTarget2D(m_textureColourTransform);
    }
```

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -196,6 +208,27 @@ bool RenderStreamSceneSelector::ValidateParameters(const RenderStreamLink::Remot
         return false;
     }
 
+    const URenderStreamSettings* settings = GetDefault<URenderStreamSettings>();
+    if (settings->OCIOConfig.bIsEnabled && settings->OCIOConfig.ColorConfiguration.ConfigurationSource)
```
DevTGHa, line 216: is `settings` guaranteed to be not `nullptr`?
Otherwise, should do null check.

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -196,6 +208,27 @@ bool RenderStreamSceneSelector::ValidateParameters(const RenderStreamLink::Remot
...
+    {
+        isColourConfigurationEnabled = true;
+
+        m_colourConversionSettings.ConfigurationSource = settings->OCIOConfig.ColorConfiguration.ConfigurationSource;
+
+        if (settings->OCIOConfig.ColorConfiguration.DestinationColorSpace.ColorSpaceIndex != INDEX_NONE)
```
DevTGHa, line 227: Maybe adding comment like below will be helpful for future:
```cpp
// OCIO has two different types for colour space, ColorSpace and DisplayView.
// Colour transform of ColorSpace can be supported in any order; however, 
// DisplayView cannot be used as source.
// To use DisplayView as source, `DisplayViewDirection` needs to be set to `Inverse` with
// `DestinationDisplayView` being `DestinationDisplayView`.
// `Support inverse view transform` setting under `Plugins - OpenColorIO` in the UE project must be enabled: DSOF-31260
```

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -506,6 +539,74 @@ void RenderStreamSceneSelector::ApplyParameters(uint32_t sceneId, const TArray<A
     m_floatValuesLast = floatValues; // event parameters need to lookup previous values
 }
 
+void RenderStreamSceneSelector::GetTextureParameter(const FString& toggle, const RenderStreamLink::ImageFrameData& frameData, size_t iImage, UTextureRenderTarget2D* Texture)
```
DevTGHa, line 551: I guess this is just extracted function?

Source/RenderStream/Public/RenderStreamSceneSelector.h
```diff
@@ -38,9 +38,15 @@ class RENDERSTREAM_API RenderStreamSceneSelector
...
+    void GetTextureParameter(const FString& toggle, const RenderStreamLink::ImageFrameData& frameData, size_t iImage, UTextureRenderTarget2D* Texture);
     
     TMap<uint64_t /*id*/, RenderStreamLink::FSkeletalLayout> m_skeletalLayoutCache;
     std::vector<uint8_t> m_schemaMem;
     RenderStreamLink::ScopedSchema m_defaultSchema;
     std::vector<float> m_floatValuesLast;
+
+    UTextureRenderTarget2D* m_textureColourTransform;
+    mutable FOpenColorIOColorConversionSettings m_colourConversionSettings;
+    mutable bool isColourConfigurationEnabled = false;
```
DevTGHa: Minor suggestion:
 
```suggestion
    mutable bool m_isColourConfigurationEnabled = false;
```

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -653,94 +754,115 @@ void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash,
...
                     {
-                        return;
+                        m_textureColourTransform = UKismetRenderingLibrary::CreateRenderTarget2D(
+                            Root,
+                            Texture->SizeX,
+                            Texture->SizeY,
+                            Texture->RenderTargetFormat
+                        );
+
+                        m_textureColourTransform->SRGB = 0;
```
DevTGHa: Please add comment explaining:
```cpp
m_textureColourTransform->SRGB = 0;
m_textureColourTransform->bForceLinearGamma = true;
```

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -653,94 +754,115 @@ void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash,
...
-                    void* resource = rtResource->TextureRHI->GetNativeResource();
 
-                    RenderStreamLink::SenderFrame data = {};
-                    if (toggle == "D3D11")
+                    if (Texture->OverrideFormat != EPixelFormat::PF_FloatRGBA)
                     {
-                        data.type = RenderStreamLink::SenderFrameType::RS_FRAMETYPE_DX11_TEXTURE;
-                        data.dx11.resource = static_cast<ID3D11Resource*>(resource);
-                        auto err = RenderStreamLink::instance().rs_getFrameImage2(frameData.imageId, &data);
+                        Texture->InitCustomFormat(frameData.width, frameData.height, EPixelFormat::PF_FloatRGBA, false);
```
DevTGHa, line 805: @josh-disguise it seems like when we enable OCIO, we now use `EPixelFormat::PF_FloatRGBA` as a `Texture->OverrideFormat` so we don't get clamped pixel value during colour transform and texture copy. This makes sense for HDR pipeline, but will have some impact on SDR under OCIO. It'd be nice to hear your opinion regarding this one.

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
josharms-disguise: Oh I didn't see that it is being called. Disregard

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
DevTGHa: The high level view of order of render event is:
1. ColourTransform from `Texture` to `m_textureColourTransform`
    - fine as long as `PF_FloatRGBA` format
3. Copy back from `m_textureColourTransform` to `Texture`
    - clamp happens if `Texture` format stayed same as before.

DevTGHa (changes requested):
Overall makes sense.
Please release `m_textureColourTransform` in destructor.

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
egorovaanastasiia: It seems like the settings is created as soon as the plugin is added. Also, other usages of `URenderStreamSettings` in the codebase don't do the `nullptr` check either. So I think it's safe to skip it

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
egorovaanastasiia: Yup, just an extracted function since we use it for both colour transform and non-colour transform cases.

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
egorovaanastasiia: I still have this comment for `Texture` object a bit earlier in the code. Do you think we need to duplicate it for `m_textureColourTransform` as well?

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
egorovaanastasiia: No, I meant to resize the Texture, similarly to as we do in the original case. I call resizeTarget for m_textureColourTransform a bit later in the same function

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
egorovaanastasiia: Yup, good point, thanks

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
egorovaanastasiia: Good idea

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
DevTGHa: I see, it's fine if that's the case

DevTGHa (approved):

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -653,94 +760,115 @@ void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash,
             UObject* o = ObjectProperty->GetObjectPropertyValue(ObjectAddress);
             if (UTextureRenderTarget2D* Texture = Cast<UTextureRenderTarget2D>(o))
             {
+                if (iImage >= nImageVals)
+                {
+                    UE_LOG(LogRenderStream, Verbose, TEXT("Attempt to read a image value from disguise that is out of range. Does the metadata need to be regenerated?"));
```
ivaylo-disguise:
```suggestion
                    UE_LOG(LogRenderStream, Verbose, TEXT("Attempt to read a image value from disguise that is out of range. Does the json schema need to be regenerated?"));
```

josharms-disguise (approved):
lgtm

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -506,6 +545,74 @@ void RenderStreamSceneSelector::ApplyParameters(uint32_t sceneId, const TArray<A
...
+                FVulkanTexture* VulkanTexture = ResourceCast(rtResource->TextureRHI->GetTexture2D());
+                auto point2 = VulkanTexture->GetSizeXY();
+
+                data.type = RenderStreamLink::SenderFrameType::RS_FRAMETYPE_VULKAN_TEXTURE;
+                data.vk.memory = VulkanTexture->GetAllocationHandle();
+                data.vk.size = VulkanTexture->GetAllocationOffset() + VulkanTexture->GetMemorySize();
+                data.vk.format = frameData.format;
+                data.vk.width = uint32_t(point2.X);
+                data.vk.height = uint32_t(point2.Y);
+                // TODO: semaphores
```
pkevin-disguise, line 603: is that semaphore still required? if not maybe the comment could be removed?

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
pkevin-disguise: typo: "an image"

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -506,6 +545,74 @@ void RenderStreamSceneSelector::ApplyParameters(uint32_t sceneId, const TArray<A
     m_floatValuesLast = floatValues; // event parameters need to lookup previous values
 }
 
+void RenderStreamSceneSelector::GetTextureParameter(const FString& toggle, const RenderStreamLink::ImageFrameData& frameData, size_t iImage, UTextureRenderTarget2D* Texture)
+{
+    ENQUEUE_RENDER_COMMAND(GetTex)(
+        [Texture, toggle, frameData, iImage](FRHICommandListImmediate& RHICmdList)
+        {
+            SCOPED_DRAW_EVENTF(RHICmdList, MediaCapture, TEXT("RS Tex Param Block %d"), iImage);
```
pkevin-disguise: I believe the text will show in Unreal Insight? Kind of prefer not having abbreviations in them.

Not a blocker or anything I think.

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
josh-disguise: I don't think Unreal has an internal SDR pipeline to speak of, internally it's all HDR until the final display pipeline (standard or OCIO pipeline). Forcing float texture should be fine unless there's something I'm missing?

pkevin-disguise (approved):
Looks fine to me, only a couple of suggestion. Nothing major.

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
egorovaanastasiia: If I'm not mistaken, Taegyun asked about the potential overhead in case we use SDR texture params under OCIO. In this PR I introduced the additional steps when receiving texture params: creating an additional texture with a different format, copying textures and colour transformation. We need to have it for HDR colors support, but not for SDR ones

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
egorovaanastasiia: I actually have no idea about this one. I just extracted this method as it is. I don't think it's related to this ticket. Maybe worth creating a new one for investigation?

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
josh-disguise: SDR in this case would be when no OCIO transform is used?

egorovaanastasiia: Not quite. SDR would be when OCIO is enabled, but the destination colour space is set to SDR

josh-disguise: How come there's no requirement for a staging texture in that case?

egorovaanastasiia: We do use a staging texture in all HDR and SDR cases for colour transform. The only case we don't use it is when OCIO is disabled. (If I understood you question correctly, and by staging texture you mean the new texture with the `PF_FloatRGBA` format )

egorovaanastasiia:
**Update:**
While testing, @elaineqiu-cmd found an issue when using several texture parameters. I was able to fix it by introducing a vector of  renderTarget helpers (one for each texture parameter.) In addition, I found that we don't actually need an extra step of copying the original texture to a helper renderTarget, so I cleaned that up.

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -803,49 +809,22 @@ void RenderStreamSceneSelector::ApplyParameters(AActor* Root, uint64_t specHash,
...
-                        Texture->ResizeTarget(frameData.width, frameData.height);
-                    }
-
-                    FTextureRenderTargetResource* SourceRes = Texture->GameThread_GetRenderTargetResource();
-                    FTextureRenderTargetResource* DestRes = m_textureColourTransform->GameThread_GetRenderTargetResource();
-                    if (!SourceRes || !DestRes)
-                    {
-                        UE_LOG(LogRenderStream, Error, TEXT("Cannot get render target resource for a source or a destination"));
-                        continue;
+                        textureColourTransform->ResizeTarget(frameData.width, frameData.height);
```
pkevin-disguise, line 819: do we need to resize all the time when we get here or could we do it only if the size is different?

egorovaanastasiia: I hadn't changed it initially since it was the same way we did for original Texture, but you're probably right. I updated it for all cases to resize only if one measurement is incorrect,

josharms-disguise (approved):
New changes lgtm

pkevin-disguise (approved):
lgtm

PR #120, opened 2026-02-03
Colours Look Incorrect When using OCIO with UE5.6+

josharms-disguise:
[RSP-376](https://d3technologies.atlassian.net/browse/RSP-376)

### **Technical Description** 

When using OCIO with Unreal Engine 5.6 or later the resulting image that is displayed in d3 looks very off. The colours appear washed out and this is a result of a change Epic made to the engine in 5.6 that caused 2 different issues.

**1. Incorrect Gamma Before OCIO Transform**

OCIO does gamma correction before the actual transform. In nDisplay there is a value called EngineDisplayGamma that is being set to 1.0 while in 5.5 and previous versions it was being set to 2.2. There is a later calculation that OCIO performs to find the gamma correction using this value which results in the wrong gamma being calculated before the transform altering the pixels to look so different on the output.

**2. Unwanted Encoding Conversion**

When nDisplay tries to copy the frame to the InputShaderResoure texture there is a check to make sure the colour encodings are the same. Since EngineDisplayGamma = 1.0 it reports that the internal render target has a Linear encoding, the InputShaderResource on the other hand has a Gamma encoding. This results in a another colour conversion that results in the image looking brighter than intended

### **Solution**

The fix in this case is to just perform the OCIO transform manually since we cannot modify any of the required variables from the plugin side. Unreal already has an API that allows us to add another pass to the render pipeline specifically to perform an OCIO transform. The other trick is to make sure the nDisplay doesn't do any colour modification behind the scenes either. Normally if nDisplay doesn't receive any colour configuration it will apply some tone curve adjustment. Luckily Unreal also provides the CaptureLateOCIO flag that we can pass to nDisplay to inform it not to do any sort of colour modification as we will be using OCIO manually at a later point. Even with this manual work the second bug still lingers as that conversion happens behind the scenes before RenderStream retrieves the final texture. To account for this we set the gamma value that is passed to the OCIO shader to 2.2. This is because the Linear->Gamma conversion that happens is pow(input, 1.0/2.2) and before the actual colour transform OCIO will perform gamma correction in the form of pow(input, gamma). So we are essentially reversing that conversion in the shader.

[RSP-376]: https://d3technologies.atlassian.net/browse/RSP-376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Enhanced OpenColorIO color management with improved resource caching and transformation workflow.
  * Integrated viewport media state updates for dynamic color handling adjustments.

* **Bug Fixes**
  * Added null reference validation to prevent crashes in media output operations.
  * Improved stability and safety of color transformation processing.


Source/RenderStream/Private/RenderStreamCapturePostProcess.cpp
```diff
@@ -109,7 +114,59 @@ void FRenderStreamCapturePostProcess::PerformPostProcessViewAfterWarpBlend_Rende
             return;
         }
 
-        Stream->SendFrame_RenderingThread(RHICmdList, frameResponse, Resources[0], Rects[0]);
+        // Apply OCIO manually if resources are cached
+        if (Module->CachedOCIOResources_RenderThread.IsValid())
```
ivaylo-disguise: It'd be nice to have this in a separate function.

Source/RenderStream/Private/RenderStreamCapturePostProcess.cpp
```diff
@@ -109,7 +114,59 @@ void FRenderStreamCapturePostProcess::PerformPostProcessViewAfterWarpBlend_Rende
...
+        {
+            const FIntPoint OutputSize(Rects[0].Width(), Rects[0].Height());
```
ivaylo-disguise: Is the output size always the same? I'm thinking of a scenario where we're using a splitting strategy in designer that results in fragments of different sizes. If such fragments use the same instance of this callback, ResultTex might be the wrong resolution for one of the fragments.

Source/RenderStream/Private/RenderStreamCapturePostProcess.cpp
```diff
@@ -109,7 +114,59 @@ void FRenderStreamCapturePostProcess::PerformPostProcessViewAfterWarpBlend_Rende
...
+            FRDGTextureRef OutputTex = GraphBuilder.RegisterExternalTexture(
+                CreateRenderTarget(ResultTex, TEXT("RSCaptureOCIOOutput")));
+
+            FScreenPassTexture Input(InputTex, Rects[0]);
+            const FIntRect OutputRect(FIntPoint::ZeroValue, OutputSize);
+            FScreenPassRenderTarget Output(OutputTex, OutputRect, ERenderTargetLoadAction::EClear);
+
+            // In UE5.6+ nDisplay applies an unwanted transform to change the encoding from Linear->sRGB on the frame
+            // The OCIO shader applies pow(input, gamma) before the transform
+            // Setting it to 2.2 undones the encoding transform to get back the linear data
```
ivaylo-disguise: Does that mean the RS plugin always expects linear gamma? How is this affected by the "Transform Source" OCIO setting in the RS plugin settings, if we set the transform source to a non-linear gamma space?

ivaylo-disguise:
Should we target RS2.0-UE5.6 with this fix?

Source/RenderStream/Private/RenderStreamCapturePostProcess.cpp
josharms-disguise: That's a good point. I just tried it with the source colour space set to sRGB - texture and it matches what 5.5 outputs. So I am a bit confused on why that worked. They both might be wrong ?

josharms-disguise:
> Should we target RS2.0-UE5.6 with this fix?

I was going to make a separate PR targeting 5.6 after these changes were finalized. I've been doing all my work using 5.7 so that's why I put this one up first.

Source/RenderStream/Private/RenderStreamCapturePostProcess.cpp
josharms-disguise: The output size is determined by the size of the view port that it's rendering to so each fragment could have a different output size. Also we use the viewport id to separate the textures so it won't constantly be getting overwritten. To be fair though I haven't tested multiple fragments on one node.

Source/RenderStream/Private/RenderStreamCapturePostProcess.cpp
pkevin-disguise: maybe the implicit transform done by unreal is always linear 🤔

Source/RenderStream/Private/RenderStream.cpp
```diff
@@ -784,26 +788,48 @@ void FRenderStreamModule::OnBeginFrame()
...
-        RootActor->GetConfigData()->StageSettings.ViewportOCIO.AllViewportsOCIOConfiguration.ColorConfiguration = settings->OCIOConfig.ColorConfiguration;
+        // Feature level refers to the shader model (SM5, SM6, etc.)
+        const ERHIFeatureLevel::Type FeatureLevel = GWorld->Scene->GetFeatureLevel();
+        FOpenColorIORenderPassResources Resources = FOpenColorIORendering::GetRenderPassResources(
+            settings->OCIOConfig.ColorConfiguration, 
+            FeatureLevel);
+
+        CachedOCIOResources = Resources;
+        CachedOCIOFeatureLevel = FeatureLevel;
     }
```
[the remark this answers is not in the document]
pkevin-disguise: wait did the a.i. just cook there? 😲

Source/RenderStream/Private/RenderStreamCapturePostProcess.cpp
```diff
@@ -164,9 +164,10 @@ void FRenderStreamCapturePostProcess::ApplyOCIOTransform(FRHICommandListImmediat
     OutputRect = FIntRect(FIntPoint::ZeroValue, OutputSize);
     FScreenPassRenderTarget Output(ShaderOutput, OutputRect, ERenderTargetLoadAction::EClear);
 
-    // In UE5.6+ nDisplay applies an unwanted transform to change the encoding from Linear->sRGB on the frame
-    // The OCIO shader applies pow(input, gamma) before the transform
-    // Setting it to 2.2 undoes the encoding transform to get back the linear data
+    // In UE5.6+ if r.DefaultBackBufferPixelFormat=3 then the EngineDisplayGamma is set to 1.0, which causes InternalRenderTarget to get a linear encoding
+    // This a mismatch from the InputShaderResource  which it resolves to. It does a conversion (pow(input, 1/2.2)) before the copy
+    // The OCIO shader applies pow(input, gamma) before the transform so we make it 2.2 to undo the previous transform
+    // r.DefaultBackBufferPixelFormat is hardcoded in d3 to always be set to 3 so we don't need to worry about supporting other values
```
pkevin-disguise, line 170: both for compressed and uncompressed?

josharms-disguise: No matter what the transport type is this always gets set to 3. However I did just realize that isn't the case for packaged projects. We do any early return if it's an exe before setting any ini variables. So there could be potential issues with Packaged UE5.6+ w/ OCIO. Could be a separate ticket though.

Source/RenderStream/Private/RenderStream.cpp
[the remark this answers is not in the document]
josharms-disguise: Nice one rabbit. I put the assignment on the render thread. I opted not to clear the cache since the module is recreated every time we start a workload. We also currently have no way to enable/disable OCIO at runtime.

pkevin-disguise (approved):
lgtm

Source/RenderStream/Private/RenderStream.cpp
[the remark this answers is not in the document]
DevTGHa: @coderabbitai review the updated diff using the conversation in this thread.

Source/RenderStream/Private/RenderStream.cpp
```diff
@@ -784,26 +788,53 @@ void FRenderStreamModule::OnBeginFrame()
...
 }
 
+// Triggered by nDisplay at the start of each frame
+void FRenderStreamModule::OnUpdateViewportMediaState(IDisplayClusterViewport* InViewport, EDisplayClusterViewportMediaState& InOutMediaState)
+{
+    // Since we are no longer setting the viewport OCIO config, nDisplay no longer disables tonemapper automatically
+    // Need to explicitly tell nDisplay that we want to apply OCIO independently at a later point
+    const URenderStreamSettings* Settings = GetDefault<URenderStreamSettings>();
+    if (Settings->OCIOConfig.ColorConfiguration.ConfigurationSource != nullptr)
```
DevTGHa, line 820: Do we need null check for `Settings` here? Or is it guaranteed?

ivaylo-disguise:
lgtm

Source/RenderStream/Private/RenderStream.cpp
josharms-disguise: It's guaranteed. Unreal always creates a default object for it first, never returns null

pkevin-disguise (commented):
this should target 5.6 and merge the fix up.

josharms-disguise:
> this should target 5.6 and merge the fix up.

Now targeting 5.6

ivaylo-disguise (approved):

DevTGHa (approved):

PR #121, opened 2026-02-05
OCIO Is Enabled tick box doesn't actually disable OCIO when unticked

josharms-disguise:
[RSP-383](https://d3technologies.atlassian.net/browse/RSP-383)

### **Technical Description**

To check if OCIO is enabled the plugin checks if there is a valid config available. So when the tick box is unticked, OCIO is still enabled because the config is not deleted.

### **Solution**

Actually set the OCIO nDisplay setting to be disabled. Check the RenderStream bIsEnabled setting rather than just relying on if there is a valid config.

[RSP-383]: https://d3technologies.atlassian.net/browse/RSP-383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStream/Private/RenderStream.cpp
+++ b/Source/RenderStream/Private/RenderStream.cpp
@@ -781,10 +781,13 @@ void FRenderStreamModule::OnBeginFrame()
     const URenderStreamSettings* settings = GetDefault<URenderStreamSettings>();
 
     ADisplayClusterRootActor* const RootActor = IDisplayCluster::Get().GetGameMgr()->GetRootActor();
-    if (RootActor && settings->OCIOConfig.ColorConfiguration.ConfigurationSource != nullptr)
+    if (RootActor)
     {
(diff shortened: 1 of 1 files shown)


josharms-disguise:
This same fix will be applied to 5.3 & 5.4 once it's approved. However once this PR (https://github.com/disguise-one/RenderStream-UE/pull/120) goes through for 5.6 & 5.7  the logic will need to be slightly different

pkevin-disguise (commented):
The ticket mentions it being reproducible on 5.3. We should fix it there and merge up.

ivaylo-disguise (approved):
looks good

josharms-disguise:
> The ticket mentions it being reproducible on 5.3. We should fix it there and merge up.

Targetting 5.3 now

pkevin-disguise (approved):
lgtm

elaineqiu-cmd (approved):
QA passed.

PR #125, opened 2026-02-19
RSP-383 5.6 Version

josharms-disguise:
[RSP-383](https://d3technologies.atlassian.net/browse/RSP-383)

### Technical Description

When unticking the enable tick box in the RenderStream OCIO settings it doesn't actually disable OCIO. This is because the only thing the plugin checks to see if it should use OCIO is if there is a valid config. This does not get deleted when disabling OCIO leading it being always on. With the OCIO workaround introduced in 5.6 the logic for this and later versions are slightly different. Since we don't worry about setting the nDisplay OCIO config anymore the logic becomes simpler. Instead all we need to do to enforce the enable tick box setting is to add a check for bIsEnabled before caching any OCIO resources.

[RSP-383]: https://d3technologies.atlassian.net/browse/RSP-383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# RS2.0-UE5.3-v8
+Compatible with disguise designer version r25.0 and above.
+* Fixed compatibility issue with the SonyCameraAndDisplay plugin
+
+# RS2.0-UE5.3-v7
+Compatible with disguise designer version r25.0 and above.
+* Fixed issue with custom events not being triggered when used in sublevels
+
 # RS2.0-UE5.3-v6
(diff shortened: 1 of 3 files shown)


ivaylo-disguise (approved):
lgtm

pkevin-disguise (approved):
lgtm

egorovaanastasiia (approved):
lgtm

PR #129, opened 2026-03-24
RS Local texture parameters latency improvements

egorovaanastasiia:
[DSOF-30790](https://d3technologies.atlassian.net/browse/DSOF-30790)

**Description:**
Texture parameters for RSL were reported to have a high latency, around 5-9 frames.

This PR consists of d3 changes in https://github.com/disguise-one/d3/pull/5933 and plugin changes in this PR.

There are 2 main changes in plugin:

- Creating new `rs_registerTextureParams()` API endpoint to read shared texture before using it. Must be called from the UE render thread.
- In case of running old d3 version + RS3.0 UE plugin , we'd like to log an error message explaining why workload has crashed. To achieve it, I made some changes to `RenderStream::StartupModule()` logic, so we cache the error, load `rs_logToD3` and show the message in d3before closing UE. 

[DSOF-30790]: https://d3technologies.atlassian.net/browse/DSOF-30790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStream/Private/RenderStream.cpp
+++ b/Source/RenderStream/Private/RenderStream.cpp
@@ -183,16 +183,38 @@ void FRenderStreamModule::StartupModule()
 
     // This code will execute after your module is loaded into memory; the exact timing is specified in the .uplugin file per-module
 
-    if (!RenderStreamLink::instance().loadExplicit())
+    FString loadError;
+    if (!RenderStreamLink::instance().loadExplicit(loadError))
     {
-        UE_LOG(LogRenderStream, Error, TEXT ("Failed to load RenderStream DLL - d3 not installed?"));
+        if (loadError.IsEmpty())
+        {
+            UE_LOG(LogRenderStream, Error, TEXT("Failed to load RenderStream DLL - d3 not installed?"));
+        }
+        else
+        {
+            if (RenderStreamLink::instance().rs_initialise)
+            {
+                int errCode = RenderStreamLink::instance().rs_initialise(RENDER_STREAM_VERSION_MAJOR, RENDER_STREAM_VERSION_MINOR);
+                if (errCode == RenderStreamLink::RS_ERROR_SUCCESS)
+                {
+                    FString msg = FString::Printf(TEXT("Failed to load %s from RenderStream DLL - d3 version is incompatible with this plugin version. Please update d3 or use an older plugin version"), *loadError);
+                    if (RenderStreamLink::instance().rs_logToD3)
+                        RenderStreamLink::instance().rs_logToD3(TCHAR_TO_ANSI(*msg));
+                }
+            }
+            if (!GIsEditor)
+            {
+                FPlatformMisc::RequestExit(true);
+            }
+            return;
+        }
     }
     else
     {
         m_logDevice = MakeShared<FRenderStreamLogOutputDevice, ESPMode::ThreadSafe>();
-        
+
         int errCode = RenderStreamLink::instance().rs_initialise(RENDER_STREAM_VERSION_MAJOR, RENDER_STREAM_VERSION_MINOR);
-        
+
         if (errCode != RenderStreamLink::RS_ERROR_SUCCESS)
         {
             if (errCode == RenderStreamLink::RS_ERROR_INCOMPATIBLE_VERSION)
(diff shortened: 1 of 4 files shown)


pkevin-disguise:
in the description, we have a case for old version of d3 + the new plugin. what about the new version of d3 + old plugin case?

egorovaanastasiia:
> in the description, we have a case for old version of d3 + the new plugin. what about the new version of d3 + old plugin case?

Sorry, mentioned it in the [other PR](https://github.com/disguise-one/d3/pull/5933) but not here: newer d3 + old plugin would still work, just with no improvements for "Waiting in cache"

Source/RenderStream/Private/RenderStream.cpp
```diff
@@ -183,16 +183,38 @@ void FRenderStreamModule::StartupModule()
...
+                if (errCode == RenderStreamLink::RS_ERROR_SUCCESS)
+                {
+                    FString msg = FString::Printf(TEXT("Failed to load %s from RenderStream DLL - d3 version is incompatible with this plugin version. Please update d3 or use an older plugin version"), *loadError);
+                    if (RenderStreamLink::instance().rs_logToD3)
+                        RenderStreamLink::instance().rs_logToD3(TCHAR_TO_ANSI(*msg));
+                }
+            }
+            if (!GIsEditor)
+            {
+                FPlatformMisc::RequestExit(true);
```
ivaylo-disguise, line 207: I wonder if this will suffer from the same exit infinite loop issue as the one seen in the UE ghost process problem. Try this in an nDisplay session with 2 RXs.

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -526,6 +527,13 @@ void RenderStreamSceneSelector::ApplyParameters(uint32_t sceneId, const TArray<A
         UE_LOG(LogRenderStream, Error, TEXT("Unable to get float frame parameters - %d"), res);
         return;
     }
+    if (nImageParams > 0)
+    {
+        ENQUEUE_RENDER_COMMAND(RegisterTextureParams)([](FRHICommandListImmediate& RHICmdList) {
+            RenderStreamLink::instance().rs_registerTextureParams();
+        });
+        FlushRenderingCommands();
```
ivaylo-disguise, line 535: I'm worried about the performance impact of synchronizing the GPU and CPU threads with the flush. If we really need the flush, please check performance is not impacted too much.

Source/RenderStream/Private/RenderStream.cpp
egorovaanastasiia: No, both RXs in this case kill the UE process. All good

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
egorovaanastasiia: ~You might be right. Seems like with a heavier scene with 9 TPs the fps of the streams dropped by 2-3 frames~
Actually, after proper retesting, seems like the performance is very similar. Version with changes takes ~10 second to stabilize, but at the end the stream framerate seems very close. Tested on the same asset, same TPs, same target resolution.
This is RS2.0:
RS2 0_larger_res
This is RS3.0:
RS3 0_larger_res

Source/RenderStream/Private/RenderStreamLink.cpp
```diff
@@ -74,7 +74,8 @@ namespace {
 
 RenderStreamLink::RenderStreamLink()
 {
-    loadExplicit();
+    FString unused;
```
josharms-disguise: Unused ? Why do we need this ?

egorovaanastasiia: Sorry, just noticed your comment.

I changed the function's `RenderStreamLink::loadExplicit(FString& outError)` signature, not it has a string param. We use it in when we call it from `StartupModule()` to log the errors, but we don't really use  it here, so Claude suggested to name it "unused".

pkevin-disguise: unused doesn't really say what it is though.

egorovaanastasiia: Ok, does "outError" sound good enough?

ivaylo-disguise (approved):
lgtm

Source/RenderStream/Private/RenderStreamLink.cpp
pkevin-disguise: yea,  it would be better.

egorovaanastasiia: Updated

pkevin-disguise (approved):
lgtm

josharms-disguise (approved):
LGTM

egorovaanastasiia:
Closing in favour of https://github.com/disguise-one/RenderStream-UE/pull/140

PR #130, opened 2026-04-08
RSP-391: Optimize idle GPU usage when not receiving frame requests

ivaylo-disguise:
Skip scene rendering, capture pipeline, and catch-up ticks when disguise is not requesting frames, freeing GPU resources for other engines (e.g. Notch) sharing the same render nodes.

- Disable world rendering via bDisableWorldRendering when m_frameDataValid is false (main GPU savings)
- Early return in post-process when no frame response exists for the current render frame (skips blit/flush/sendFrame)
UE_idle_workload_optimizations

### DIFF
--- a/Source/RenderStream/Private/RenderStream.cpp
+++ b/Source/RenderStream/Private/RenderStream.cpp
@@ -781,6 +781,11 @@ void FRenderStreamModule::OnBeginFrame()
     if (IsController)
         m_syncFrame.ControllerReceive();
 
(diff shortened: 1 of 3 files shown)


pkevin-disguise (approved):
lgtm

egorovaanastasiia (approved):
looks good!

josharms-disguise (approved):
LGTM another bonus for the 3.0 release

elaineqiu-cmd (approved):
QA passed.

PR #131, opened 2026-04-17
RSP-393: Duplicated camera names can cause inconsistent camera selection within cluster

ivaylo-disguise:
### Technical Description:

When multiple cameras share the same RenderStream channel name (either within a single level or across streaming sub-levels), the camera selected at runtime was determined by BeginPlay() execution order. This order is non-deterministic across nDisplay cluster nodes, meaning different machines in the cluster could pick different cameras for the same channel, leading to inconsistent rendering.

### Solution:

Three changes were made:

Deterministic camera selection — Instead of using the last-registered camera (Array.Last()), the selection now sorts candidates by preferring cameras in the persistent level first, then tie-breaking by GetPathName() (which is stable across machines since actor/level paths are serialized in the packaged project).
Runtime warning — A log warning is emitted during BeginPlay() when a second camera registers on an already-occupied channel, alerting users to the duplicate.
Editor-time validation — Two new checks in the editor detect duplicate channel names: one within the same level (in UpdateLevelChannelCache) and one across different levels (in FRenderStreamValidation::RunValidation). Both report errors through the RenderStreamValidation message log.

### Affected areas:

RenderStreamChannelDefinition — camera selection logic and BeginPlay registration
RenderStreamEditorModule — level channel cache building
RenderStreamValidation — cross-level validation pass

### DIFF
--- a/Source/RenderStream/Private/RenderStreamChannelDefinition.cpp
+++ b/Source/RenderStream/Private/RenderStreamChannelDefinition.cpp
@@ -94,6 +94,34 @@ uint32 URenderStreamChannelDefinition::GetChannelCameraNum(const FString& Channe
     return (*Actors)->Num();
 }
 
+inline static TWeakObjectPtr<ACameraActor> GetBetterCamera(const TWeakObjectPtr<ACameraActor>& CamA, const TWeakObjectPtr<ACameraActor>& CamB)
+{
+    // prefer valid camera 
+    if (!CamB.IsValid())
+        return CamA;
+
+    if (!CamA.IsValid())
+        return CamB;
+
+    const bool CamAIsPersistent = CamA->GetLevel() && CamA->GetLevel()->IsPersistentLevel();
+    const bool CamBIsPersistent = CamB->GetLevel() && CamB->GetLevel()->IsPersistentLevel();
+
+    if (CamAIsPersistent != CamBIsPersistent)
+    {
+        // one is persistent, the other not: prefer the camera from the persistent level
+        if (CamAIsPersistent)
+            return CamA;
+        else
+            return CamB;
+    }
+
+    // final tiebreaker: prefer smaller path
+    if (CamA->GetPathName() < CamB->GetPathName())
+        return CamA;
+    else
+        return CamB;
+}
+
 TWeakObjectPtr<ACameraActor> URenderStreamChannelDefinition::GetChannelCamera(const FString& Channel)
 {
     if (Channel.IsEmpty())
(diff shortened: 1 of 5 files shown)


Source/RenderStream/Private/RenderStreamChannelDefinition.cpp
```diff
@@ -93,6 +94,34 @@ uint32 URenderStreamChannelDefinition::GetChannelCameraNum(const FString& Channe
     return (*Actors)->Num();
 }
 
+inline static TWeakObjectPtr<ACameraActor> GetBetterCamera(const TWeakObjectPtr<ACameraActor>& CamA, const TWeakObjectPtr<ACameraActor>& CamB)
+{
+    // prefer valid camera 
+    if (!CamB.IsValid())
```
pkevin-disguise, line 100: here if CamA is also invalid, we return an invalid camera. Is that intented?

ivaylo-disguise: Yes, if both cameras are invalid it doesn't matter which one we keep in this comparison. We just hope that in the next comparison, the camera will be valid.

If all cameras are invalid, we will end up picking CamA overall. This is slightly different than the old code, which would have picked the last camera `return (*ActorsPtrPtr)->Last();`.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -546,6 +548,16 @@ URenderStreamChannelCacheAsset* UpdateLevelChannelCache(ULevel* Level)
             if (Definition.IsValid())
             {
                 FString ChannelName = TCHAR_TO_UTF8(*(Definition->GetChannelName()));
+                // Detect duplicate channel names within this level. The cross-level case
+                // is handled separately in FRenderStreamValidation::RunValidation().
+                if (Cache->Channels.Contains(ChannelName))
+                {
+                    FMessageLog RSV("RenderStreamValidation");
+                    RSV.Error()->AddToken(FTextToken::Create(FText::FromString(FString::Printf(
```
egorovaanastasiia: Do you know how often does it get logged? Could it be too spammy?

ivaylo-disguise: Only when you save the project in UnrealEditor.

Source/RenderStream/Private/RenderStreamChannelDefinition.cpp
ivaylo-disguise: I'm looking into whether it makes sense to `return FindCameraInScene();` in the case all cameras for this channel are invalid.

ivaylo-disguise: Alright, here are the conclusions:
**Current behaviour:**
- If channel name empty -> use `FindCameraInScene()`
- If no actor matches the channel name -> use `FindCameraInScene()`
- If actor matches channel name, but has been garbage collected (weak-ptr invalid) -> spam 'Failed to find camera' errors and render with default viewpoint values which will make it obvious to the user something's broken
Image

If we use the `FindCameraInScene()` when all camera actors are invalid, then we would not get the error spam and would default to the first camera found, which may be the one the user was expecting, but also may be a random camera. Either way, I think it's more confusing as it could possibly mask issues.
Image

pkevin-disguise: Agreed.

pkevin-disguise (approved):
lgtm

egorovaanastasiia (approved):
lgtm

josharms-disguise (approved):
LGTM

egorovaanastasiia (approved):
lgtm

elaineqiu-cmd (approved):
QA passed.

PR #138, opened 2026-05-20
Support Array Parameter Type

josharms-disguise:
[RSP-385](https://d3technologies.atlassian.net/browse/RSP-385?atlOrigin=eyJpIjoiMzI1MDVkNTBkNzM4NDRhNzkwOThhZTk0MmVhNzQ0MjciLCJwIjoiaiJ9)

### Summary

D3 already supports array parameters when using notch. This change allows float array parameters created in the blueprint to be exposed to d3. Since arrays need to be handled differently a new type is used instead of try to represent it as a number. It essentially follows the same pattern as the other types.

[RSP-385]: https://d3technologies.atlassian.net/browse/RSP-385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
+++ b/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
@@ -198,7 +198,15 @@ void CreateField(FRenderStreamExposedParameterEntry& parameter, FString group, F
 {
     check(type != RenderStreamParameterType::Float);
     check(type != RenderStreamParameterType::Text);
+    check(type != RenderStreamParameterType::Array);
     CreateFieldInternal(parameter, group, displayName_, suffix, key_, undecoratedSuffix, type, 0, 0, 0, "");
+}
+
+void CreateField(FRenderStreamExposedParameterEntry& parameter, FString group, FString displayName_, FString suffix, FString key_, FString undecoratedSuffix, RenderStreamParameterType type, float min, float max, float step, float defaultValue, uint32 nElements)
+{
+    check(type == RenderStreamParameterType::Array);
+    CreateFieldInternal(parameter, group, displayName_, suffix, key_, undecoratedSuffix, type, min, max, step, FString::SanitizeFloat(defaultValue));
+    parameter.NumElements = nElements;
 }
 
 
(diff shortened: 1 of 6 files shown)


Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -406,6 +407,32 @@ void GenerateParameters(TArray<FRenderStreamExposedParameterEntry>& Parameters,
...
+                    UE_LOG(LogRenderStreamEditor, Warning, TEXT("Skipping float array property %s: array is empty - size the array in the editor before exposing it"), *Name);
+                }
+                else
+                {
+                    UE_LOG(LogRenderStreamEditor, Log, TEXT("Exposed float array property: %s with %u elements"), *Name, nElements);
+                    const bool HasLimits = Property->HasMetaData("ClampMin") && Property->HasMetaData("ClampMax");
+                    const float Min = HasLimits ? FCString::Atof(*Property->GetMetaData("ClampMin")) : -1;
+                    const float Max = HasLimits ? FCString::Atof(*Property->GetMetaData("ClampMax")) : +1;
+                    FRenderStreamExposedParameterEntry& Entry = Parameters.Emplace_GetRef();
+                    CreateFieldInternal(Entry, Category, Name, "", Name, "", RenderStreamParameterType::Array, Min, Max, 0.001f, "0");
```
ivaylo-disguise: You could create another `CreateField()` overload for arrays that also takes in nElements, to keep the pattern of calling `CreateField()` instead of calling `CreateFieldInternal()` directly.

ivaylo-disguise (approved):
looks good

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
josharms-disguise: Yes good point

elaineqiu-cmd (approved):
QA passed.

PR #142, opened 2026-06-08
Restores automatic regeneration of the schema on save

egorovaanastasiia:
[RSP-397](https://d3technologies.atlassian.net/browse/RSP-397)

**Description:**
The editor module regenerated the schema by listening to `FEditorDelegates::PostSaveExternalActors()` and
`PostSaveWorldWithContext()`. After upgrading to UE 5.6 these delegates no longer fire reliably on a plain save, since UE5.6 only write packages that are actually dirty. As a result, the schema went stale until the user happened to modify an object in the scene. 

**Solution:**
 Intercept the Save command instead of listening for a save: `RegisterSaveCommandOverrides()` wraps the level editor's
 `Save `and `SaveAllLevels` actions in the global command list, so it preserves the original CanExecute/visibility, runs the
 original save, then sets `DirtyAssetMetadata = true` regardless of whether any package was written.

In addition, I removed the stale delegates.

[RSP-397]: https://d3technologies.atlassian.net/browse/RSP-397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
+++ b/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
@@ -49,6 +49,10 @@
 #include "DesktopPlatformModule.h"
 #include "IDesktopPlatform.h"
 
+#include "LevelEditor.h"
+#include "LevelEditorActions.h"
+#include "Framework/Commands/UICommandList.h"
+
 DEFINE_LOG_CATEGORY(LogRenderStreamEditor);
 
 #define LOCTEXT_NAMESPACE "RenderStreamEditor"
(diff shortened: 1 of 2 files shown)


josharms-disguise (approved):
LGTM

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -1002,15 +1005,40 @@ void FRenderStreamEditorModule::RegisterToolBarButton()
...
+                OriginalExecute.Execute();
+
+            // Mark dirty so OnBeginFrame regenerates next tick, regardless of whether a package was written.
+            DirtyAssetMetadata = true;
+        });
+
+        CommandList->MapAction(Command, WrappedAction);
+    };
+
+    WrapSaveCommand(FLevelEditorCommands::Get().Save);
```
ivaylo-disguise, line 1040: If you think there's a case where the RenderStream plugin can get unloaded at runtime, we should restore the original save commands to avoid calling into an invalid instance of the plugin.

ivaylo-disguise: Looks like this can't happen.

ivaylo-disguise (approved):

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
egorovaanastasiia: I tried reproing the case that Ivaylo suggested:

- disable the plugin while workload is running  
- UE editor asks to restart the session 
- after restarting, plugin is disabled, workload is still running
- after that if workload is restarted, it'll crash
Both my plugin fix and original plugin behave the same way in this case.

I also tried resaving the asset while workload is running, I noticed no issues there.

pkevin-disguise (approved):
lgtm

elaineqiu-cmd (approved):
QA passed.

PR #143, opened 2026-06-10
Error message in schema generation when using old incompatible d3 version

egorovaanastasiia:
[RSP-401](https://d3technologies.atlassian.net/browse/RSP-401)

**Description:**
Old d3 version and plugin RS3.0-UE5.x are not backward compatible. This caused the schema generation to fail, causing an accesses violation that would be only seen in Output Logs of UE editor (which can be missed easily.)

Solution was to gate schema generation, so before calling `rs_saveSchema`, we read the actual `d3renderstream.dll` file version and refuse generation with a clear actionable message, which also points to the first supported d3 version. The error looks like this in the UE editor:
image

[RSP-401]: https://d3technologies.atlassian.net/browse/RSP-401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStream/Private/RenderStreamLink.cpp
+++ b/Source/RenderStream/Private/RenderStreamLink.cpp
@@ -19,6 +19,7 @@
 THIRD_PARTY_INCLUDES_START
 #include <winver.h>
 THIRD_PARTY_INCLUDES_END
+#include "Windows/HideWindowsPlatformTypes.h"
 #pragma comment(lib, "version.lib")
 #endif
 
@@ -287,18 +288,7 @@ bool RenderStreamLink::IsSchemaGenerationSupported() const
         return false;
     }
 
-    const bool supported =
-        (major > MIN_D3_VERSION_MAJOR) ||
+    return (major > MIN_D3_VERSION_MAJOR) ||
         (major == MIN_D3_VERSION_MAJOR && minor > MIN_D3_VERSION_MINOR) ||
         (major == MIN_D3_VERSION_MAJOR && minor == MIN_D3_VERSION_MINOR && patch >= MIN_D3_VERSION_PATCH);
-
-    if (!supported)
-    {
-        UE_LOG(LogRenderStream, Error,
-            TEXT("Installed d3 version %d.%d.%d is too old for RenderStream schema generation; d3 version r%d.%d.%d or newer is required."),
-            major, minor, patch,
-            MIN_D3_VERSION_MAJOR, MIN_D3_VERSION_MINOR, MIN_D3_VERSION_PATCH);
-    }
-
-    return supported;
 }
--- a/Source/RenderStream/Private/RenderStreamLink.cpp
+++ b/Source/RenderStream/Private/RenderStreamLink.cpp
@@ -14,6 +14,14 @@
 #include "Interfaces/IPluginManager.h"
 #include "Windows/MinWindows.h"
 
+#ifdef WINDOWS
+#include "Windows/AllowWindowsPlatformTypes.h"
+THIRD_PARTY_INCLUDES_START
+#include <winver.h>
+THIRD_PARTY_INCLUDES_END
+#pragma comment(lib, "version.lib")
+#endif
+
 namespace {
     void log_default(const char* text) {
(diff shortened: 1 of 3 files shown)


Source/RenderStream/Public/RenderStreamLink.h
```diff
@@ -362,6 +362,10 @@ class RenderStreamLink
 #define RENDER_STREAM_VERSION_MAJOR 3
 #define RENDER_STREAM_VERSION_MINOR 0
 
+#define MIN_D3_VERSION_MAJOR 33
+#define MIN_D3_VERSION_MINOR 2
+#define MIN_D3_VERSION_PATCH 0
+
```
egorovaanastasiia, line 368: These are temporary values for now. Need to be updated when we release d3 version that would support RS3.0

Source/RenderStream/Private/RenderStreamLink.cpp
```diff
@@ -14,6 +14,14 @@
 #include "Interfaces/IPluginManager.h"
 #include "Windows/MinWindows.h"
 
+#ifdef WINDOWS
+#include "Windows/AllowWindowsPlatformTypes.h"
```
ivaylo-disguise, line 18: I think you may need to mirror the allow windows types with an `#include "Windows/HideWindowsPlatformTypes.h"` after you've included the windows headers you need.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -774,6 +774,17 @@ void FRenderStreamEditorModule::GenerateAssetMetadata()
         return;
     }
 
+    if (!RenderStreamLink::instance().IsSchemaGenerationSupported())
+    {
+        const FString msg = FString::Printf(
+            TEXT("Cannot generate RenderStream schema: the installed d3 version is unsupported or could not be determined. Schema generation requires d3 r%d.%d.%d or newer."),
+            MIN_D3_VERSION_MAJOR, MIN_D3_VERSION_MINOR, MIN_D3_VERSION_PATCH);
+        UE_LOG(LogRenderStreamEditor, Error, TEXT("%s"), *msg);
```
ivaylo-disguise, line 777: IsSchemaGenerationSupported already logged an error message for this. I think we can drop the UE_LOG here.

ivaylo-disguise (changes requested):

Source/RenderStream/Private/RenderStreamLink.cpp
egorovaanastasiia: You're right, thanks for pointing

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
egorovaanastasiia: Yup, it's double the log. I removed the log from `IsSchemaGenerationSupported()` and left this one here instead

ivaylo-disguise (approved):
lgtm

pkevin-disguise (approved):

egorovaanastasiia:
**Update:**
@elaineqiu-cmd found an issue where schema wasn't regenerated and was failing silently again. 

The problem occurred after I re-merged `RS3.0-UE5.5` branch, which had latest changes for RSL improvements. The main issue was that in `GenerateAssetMatadata()` it would first check if `RSLink` is available (which was failing since I introduced the new API endpoint for RSL ticket) and only then it would check if the d3 version is supported.

The solution was to change the order: first check the d3 version, log the message in case of an error, and only then check if `RSLink`  is available.

ivaylo-disguise (approved):

josharms-disguise (approved):
LGTM

elaineqiu-cmd (approved):
QA passed.

Source/RenderStream/Public/RenderStreamLink.h
egorovaanastasiia: The d3 release is indeed will be r33.3. No changes needed, merging this PR

PR #145, opened 2026-06-26
RSP-403 fix startup crash

ivaylo-disguise:
Compressed + Packaged build expose a timing issue. Each stream has a buffer texture which is created asynchronously on the render queue. With tighter timings, it's possible to attempt using this texture before it's initialized.

Solution: add safety by dropping the frame in SendFrame when texture note ready.

This bug was discovered when testing packaged builds in RSP-403.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability when sending rendered frames by skipping processing if the destination texture is not available.
  * Added a safe early exit and logging to help avoid crashes in this scenario.


egorovaanastasiia (approved):
LGTM

elaineqiu-cmd (approved):
QA passed.

PR #147, opened 2026-07-02
5.8 Update

josharms-disguise:
## Summary

Changes necessary to get the 5.8 plugin to work. The biggest change is that we are now copying the renderstream DLL and any related ones from the installation directory to a ThirdParty folder within the UE project folder. This is a workaround to prevent UE from parsing the unwind data from the d3renderstreamsend.dll. This occurs with every DLL loaded by Unreal but since this particular DLL is encrypted it causes a crash. Being under the ThirdParty folder exempts these DLL from building the unwind table.

## New Changes

During QA 2 new issues were found that required more code changes. First there was an issue when running a workload with more than one node, the follower node wouldn't start at all. This is because the methods SerializeToString() & DeserializeFromString were deprecated in 5.8. These functions are what nDisplay uses to send information from the controller to the followers. We override these functions to pass along RenderStream specific info. Since the follower was not getting any info it was not able to start properly. The fix is to use the new function provided: SerializeDC(). This takes the role of both deprecated functions so it encapsulates both of their behaviours. 

* Files affected: SyncFrameData.h & SyncFrameData.cpp

The second issue was that it was not possible to run 2 UE processes using the same project. This is because of the workaround to copy over the d3 DLLs to the ThirdParty folder (d3renderstream.dll, etc.). We need to guarantee that the DLLs are the same as the current d3 version which could change at any time. This causes a problem since the second process won't be able to overwrite the DLLs since they are currently being used by the first process. To fix this I have made it so each process gets their own copy of the DLLs that are copied to a folder that is named after the PID. The plugin will also attempt to delete old DLLs copies if it can. 

* File affected: RenderStreamLink.cpp


Source/RenderStream/Private/RenderStream.cpp
```diff
@@ -814,18 +811,6 @@ void FRenderStreamModule::OnBeginFrame()
...
-    // With the release of 5.7 a regression was introduced that requires offscreen workloads to have valid MediaOutputs
-    // The types are just random ones I picked
-    // Epic is aware of this and is planning to fix it in 5.7.1
-    ADisplayClusterRootActor* const RootActor = IDisplayCluster::Get().GetGameMgr()->GetRootActor();
-    if (RootActor->GetConfigData()->GetNode(ClusterMgr->GetNodeId())->MediaSettings.MediaOutputs.Num() == 0)
-    {
-        TObjectPtr<UFileMediaOutput> MediaOutput = NewObject<UFileMediaOutput>(GetTransientPackage());
-        RootActor->GetConfigData()->GetNode(ClusterMgr->GetNodeId())->MediaSettings.bEnable = true;
-        RootActor->GetConfigData()->GetNode(ClusterMgr->GetNodeId())->MediaSettings.MediaOutputs.Add({MediaOutput,});
-    }
```
egorovaanastasiia, line 828: This regression was resolved in UE5.8?

Source/RenderStream/Private/RenderStreamLink.cpp
```diff
@@ -102,13 +104,39 @@ bool RenderStreamLink::loadExplicit()
...
+
+    if (copiedRSMain && copiedRSSend)
+    {
+        AddDllDirectory(*exePath);
+        dllPath = destDir / dllName;
+        UE_LOG(LogRenderStream, Log, TEXT("Copied RenderStream DLLs"));
+    }
+    else
+    {
+        UE_LOG(LogRenderStream, Warning, TEXT("Failed to copy RenderStream DLLs"));
```
egorovaanastasiia: Should we use the verbosity level - Error here? This sounds more like an error than just a warning

Source/RenderStream/Private/RenderStreamViewportClient.cpp
```diff
@@ -268,97 +268,6 @@ class FDisplayClusterSceneViewport : public FSceneViewport
         }
     }
 
-    virtual int32 DrawStatsHUD(FCanvas* InCanvas, int32 InX, int32 InY) override
```
egorovaanastasiia, line 271: We don't need it for UE5.8?

Source/RenderStream/Public/RenderStreamLink.h
```diff
@@ -28,7 +28,7 @@ typedef uint64_t VkDeviceSize;
 typedef struct VkSemaphore_T* VkSemaphore;
 
 #define RS_PLUGIN_NAME "RenderStream-UE"
-#define RS_PLUGIN_VERSION "RS2.0-UE5.7-v3"
+#define RS_PLUGIN_VERSION "RS2.0-UE5.8-v1"
```
egorovaanastasiia: Don't we start with v0?

Source/RenderStream/Private/RenderStream.cpp
josharms-disguise: Yep, it was also causing performance issues so it's good we were able to remove it

Source/RenderStream/Public/RenderStreamLink.h
josharms-disguise: You might be right

Source/RenderStream/Private/RenderStreamViewportClient.cpp
josharms-disguise: No, this file is almost an exact copy of DisplayViewportClient with some minor modifications. We were overriding this function which doesn't exist anymore

egorovaanastasiia (approved):
lgtm

Source/RenderStream/Public/RenderStreamChannelCacheAsset.h
```diff
@@ -36,21 +36,21 @@ struct FRenderStreamExposedParameterEntry
     UPROPERTY(EditAnywhere, Category = "ExposedParameter")
     FString Key;
     UPROPERTY(EditAnywhere, Category = "ExposedParameter")
-    RenderStreamParameterType Type;
+    RenderStreamParameterType Type = RenderStreamParameterType::Float;
     UPROPERTY(EditAnywhere, Category = "ExposedParameter")
-    float Min;
+    float Min = 0.f;
```
ivaylo-disguise, line 41: Would it be a good idea to apply this initialization to UE5.6 and UE5.7 as well?

Source/RenderStream/Private/RenderStreamLink.cpp
```diff
@@ -102,13 +104,39 @@ bool RenderStreamLink::loadExplicit()
...
         return false;
     }
 
+    // Workaround for 5.8, encrypted DLLs crash when loaded because UE tries to parse the callstack unwind info which fails
+    // There is a filter that exempts DLLs within a ThirdParty folder from this parsing
+    FString dllPath = exePath + dllName;
+    const FString destDir = FPaths::ConvertRelativePathToFull(FPaths::ProjectDir() / TEXT("ThirdParty"));
+    const FString srcSend = exePath + TEXT("d3renderstreamsend.dll");
+    const FString srcVideoSend = exePath + TEXT("d3libvideosend.dll");
+    IFileManager& fileManager = IFileManager::Get();
```
ivaylo-disguise, line 116: It'd be nice if this could be generalized to work on a list of DLLs and just copy them all.

Source/RenderStream/Public/RenderStreamChannelCacheAsset.h
josharms-disguise: I did this here because it is now a warning when building 5.8, I don't think it would be necessary to do this for older versions

ivaylo-disguise (approved):
lgtm

Source/RenderStream/Private/RenderStreamLink.cpp
```diff
@@ -113,13 +113,24 @@ bool RenderStreamLink::loadExplicit()
...
     FString dllPath = exePath + dllName;
-    const FString destDir = FPaths::ConvertRelativePathToFull(FPaths::ProjectDir() / TEXT("ThirdParty"));
     IFileManager& fileManager = IFileManager::Get();
+    const FString baseDir = FPaths::ConvertRelativePathToFull(FPaths::ProjectDir() / TEXT("ThirdParty"));
+
+    // Delete DLL copies not in use by a process
+    TArray<FString> priorRunDirs;
+    fileManager.FindFiles(priorRunDirs, *(baseDir / TEXT("*")), false, true);
+    for (const FString& dir : priorRunDirs)
+        fileManager.DeleteDirectory(*(baseDir / dir), false, true);
```
egorovaanastasiia, line 127: Are we deleting everything from the ThirdParty folder? What if there are other non-disguise plugins there?

Source/RenderStream/Private/RenderStreamViewportClient.cpp
egorovaanastasiia, line 271: Why was this removed?

Source/RenderStream/Private/RenderStreamLink.cpp
josharms-disguise: We delete everything but it's a folder we create in the first place, I don't think that's where actually third party stuff is supposed to go there though.

Source/RenderStream/Private/RenderStreamViewportClient.cpp
josharms-disguise: This is a class we mostly copy paste from the parent and this method was completely removed. So there is nothing we can override.

Source/RenderStream/Private/RenderStreamLink.cpp
```diff
@@ -102,13 +104,61 @@ bool RenderStreamLink::loadExplicit()
...
     {
         UE_LOG(LogRenderStream, Error, TEXT("%s not found in %s."), *dllName, *exePath);
         return false;
     }
 
+    // Workaround for 5.8, encrypted DLLs crash when loaded because UE tries to parse the callstack unwind info which fails
+    // There is a filter that exempts DLLs within a ThirdParty folder from this parsing
+    FString dllPath = exePath + dllName;
+    IFileManager& fileManager = IFileManager::Get();
+    const FString baseDir = FPaths::ConvertRelativePathToFull(FPaths::ProjectDir() / TEXT("ThirdParty"));
```
pkevin-disguise: could we scope the dlls into `ThirdParty/RenderStream/` so that when we delete the folder we don't delete unrelated third party dlls?

Source/RenderStream/Private/RenderStreamLink.cpp
```diff
@@ -102,13 +104,61 @@ bool RenderStreamLink::loadExplicit()
...
+
+    // Delete DLL copies not in use by a process
+    TArray<FString> priorRunDirs;
+    fileManager.FindFiles(priorRunDirs, *(baseDir / TEXT("*")), false, true);
+    for (const FString& dir : priorRunDirs)
+        fileManager.DeleteDirectory(*(baseDir / dir), false, true);
```
pkevin-disguise, line 127: Since there could be more then one RS workload running on the same machine, I think there could be a race where we delete the dll of a workload that is starting before it reaches the load lib call which would lock the dlls.

Source/RenderStream/Private/RenderStreamLink.cpp
egorovaanastasiia: Hmm, ok.  I'm just concerned that the folder's name is general, and some users may use it for something else. Maybe creating another `RenderStream` folder inside `ThirdParty` would be safer? 

But the final decision is up to you!

egorovaanastasiia (approved):

Source/RenderStream/Private/RenderStreamLink.cpp
pkevin-disguise: > Are we deleting everything from the ThirdParty folder? What if there are other non-disguise plugins there?

Had the same comment. I do think putting the dll in ThirdParty/RenderStream/ would be safer.

pkevin-disguise (changes requested):

Source/RenderStream/Private/RenderStreamLink.cpp
josharms-disguise: I can add that

Source/RenderStream/Private/RenderStreamLink.cpp
josharms-disguise: Hmm ya there would be a tiny fraction of time between copying them over and actually loading them that another process could delete them. What do you think of adding some kind of time limit? So it won't be able to delete folders created X minutes ago. Or even seconds.

pkevin-disguise: I feel like time wouldn't really be a robust solution. what if we had a semaphore which ensured that this method ran once across UE processes on the same machine?

josharms-disguise: I'm not sure that a semaphore it warranted for this, might add unneeded complexity. I've found a way to check if the process is running through UE's FPlatfromProcess::IsApplicationRunning(). Since we save the DLLs under a folder named after the PID we can do a smarter deletion.

pkevin-disguise: so the dlls are copied once per process?

josharms-disguise: Ya currently they are saved into their own folders based on the PID. This way we can guarantee that the DLLs are up to date and not being used by another process.

pkevin-disguise: right. so the path we delete  `/thirdparty/rs/pid/`, not `/thirdparty/rs/`

elaineqiu-cmd (approved):
QA passed.

josharms-disguise:
Standalone release. Closing since it's not being merged.

PR #148, opened 2026-07-03
Updated d3 compatible version

egorovaanastasiia:
The next `wrap/next` was changed from r33.3 to r34.0. We need to update it as the first RS3.0 compatible d3 version.

Will merge after wrap/next will be released.


ivaylo-disguise (approved):
lgtm

PR #149, opened 2026-07-08
No Level Blueprints For Exposed Parameters

josharms-disguise:
See [RSP-404](https://d3technologies.atlassian.net/browse/RSP-404?atlOrigin=eyJpIjoiNTM5OGZlMzEyZmY4NDhkMDg4YTAyOTJmZDcwMGJhYjQiLCJwIjoiaiJ9)

This is a feature requested by the Sphere to make workflows easier. They noted how they would like to be able to use the same blueprints with different levels. This is currently difficult because any exposed parameters must be done through the level blueprint and the logic isn't easily transferable.

This solution breaks the dependency of level blueprints. Instead it introduces a new class RenderStreamBlueprint that can be inherited by a blueprint. This gives us the ability to search for the blueprints we need to extract the parameters from just by checking the type, no tags or components necessary.

Another thing that has been added is an easier way to create these special blueprints through the UI. 

[RSP-404]: https://d3technologies.atlassian.net/browse/RSP-404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -46,9 +47,13 @@ void RenderStreamSceneSelector::GetAllLevels(TArray<AActor*>& Actors, ULevel * L
 {
     if (Level)
     {
-        auto Actor = static_cast<AActor*>(Level->GetLevelScriptActor());
-        if (Actor && !Actors.Contains(Actor))
```
ivaylo-disguise, line 50: Are we in risk of getting duplicates since the removal of the `!Actors.Contains(Actor)` check?

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
egorovaanastasiia: Is it gonna work for StreamingLevels? I think it might still read parameters from `GetLevelScriptActor()`

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -46,9 +47,13 @@ void RenderStreamSceneSelector::GetAllLevels(TArray<AActor*>& Actors, ULevel * L
 {
```
ivaylo-disguise, line 47: GetAllLevels is called on every frame, and it's now itterating over all actors. If we don't anticipate RenderStreamBlueprint actors to be added dynamically, maybe this list can be cached.

josharms-disguise: Good point

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
josharms-disguise: True, I'll fix it

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
josharms-disguise: Sorry, I forgot to answer. You're right I've fixed it

Source/RenderStream/Private/SceneSelector_StreamingLevels.cpp
```diff
@@ -107,25 +114,25 @@ void SceneSelector_StreamingLevels::ApplyScene(const UWorld& World, uint32_t sce
                 if (streamingLevel->IsLevelLoaded())
                 {
                     streamingLevel->SetShouldBeVisible(true);
-
-                    ApplyParameters(sceneId, { static_cast<AActor*>(spec.persistentRoot), static_cast<AActor*>(streamingLevel->GetLevelScriptActor()) });
+                    ApplyParameters(sceneId, GetSpecActors(World, spec));
                 }
             }
-            else if (spec.streamingLevel != nullptr)
+            else
```
egorovaanastasiia, line 120: Was it intentional removal of null check?

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -42,13 +43,25 @@ RenderStreamSceneSelector::~RenderStreamSceneSelector()
...
 }
 
+void RenderStreamSceneSelector::GetActorsInLevel(TArray<AActor*>& Actors, ULevel* Level) const
+{
+    if (!Level)
+        return;
+
+    for (AActor* Actor : Level->Actors)
+    {
+        if (Actor && Actor->IsA<ARenderStreamBlueprint>() && !Actors.Contains(Actor))
```
egorovaanastasiia, line 53: It's possible to have 2 or more `RenderStreamBlueprints` in 1 project? How do we handle the case when parameters have the same name between 2 blueprints?

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
egorovaanastasiia: Thanks!

Source/RenderStream/Private/SceneSelector_StreamingLevels.cpp
josharms-disguise: Yes because it's redundant, the if() at line 99 is already checking for null

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
josharms-disguise: It's possible. Do you mean what happens when they are in the same scene or different scenes? If it's the the same scene then it just gets recorded twice in that parameter list. If it's in a different scene then it gets recorded in each scenes list. The consequence of this is that d3 sees these parameters as the same since they will all share the same key (as of now key is just the name of the parameter, not unique). I originally was trying to make it so every key could be unique so we didn't have collisions but since it is behaviour people might want/expect I didn't end up changing it. It's possible we could make things more granular in the future because right now every parameter with the same name is treated like it's the exact same instance regardless of which level it's in or whatnot.

egorovaanastasiia (approved):
lgtm

Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -42,13 +43,25 @@ RenderStreamSceneSelector::~RenderStreamSceneSelector()
     }
 }
 
+void RenderStreamSceneSelector::GetActorsInLevel(TArray<AActor*>& Actors, ULevel* Level) const
+{
+    if (!Level)
+        return;
+
+    for (AActor* Actor : Level->Actors)
```
pkevin-disguise, line 51: Is the schema is idempotent between saves if I have multiple RenderStreamBlueprints in the same level?

josharms-disguise: As in the parameter order is always the same or the scene order? The parameter order seems to always be the same in alphabetical order. The map order isn't but I believe there is already a ticket to look at that.

pkevin-disguise: I mean the parameter order. Since they are in multiple different blueprints. We still need to have the generated schema to be the same between saves.

josharms-disguise: Yep that's the same

pkevin-disguise (approved):
lgtm

elaineqiu-cmd (approved):
QA passed.

PR #151, opened 2026-07-15
When nDisplay controller is present, wait for the controller to instruct exit

josh-disguise:
Makes non-primary ndisplay nodes in the renderstream session exit cleanly.

DSOF-32277

### DIFF
--- a/Source/RenderStream/Private/SyncFrameData.cpp
+++ b/Source/RenderStream/Private/SyncFrameData.cpp
@@ -236,7 +236,17 @@ void FRenderStreamSyncFrameData::Apply() const
 void FRenderStreamSyncFrameData::QuitNow() const
 {
     TRACE_CPUPROFILER_EVENT_SCOPE(TEXT("FRenderStreamSyncFrameData::QuitNow()"));
+
     RenderStreamLink::instance().rs_setNewStatusMessage("");
+
+    IDisplayClusterClusterManager* ClusterMgr = IDisplayCluster::IsAvailable() ? IDisplayCluster::Get().GetClusterMgr() : nullptr;
+
+    if (ClusterMgr && !ClusterMgr->IsPrimary())
+    {
+        UE_LOG(LogRenderStream, Log, TEXT("RenderStream requested quit, waiting for nDisplay primary to terminate this node"));
+        return;
+    }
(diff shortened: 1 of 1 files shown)


Source/RenderStream/Private/SyncFrameData.cpp
```diff
@@ -231,7 +231,17 @@ void FRenderStreamSyncFrameData::Apply() const
...
 {
     TRACE_CPUPROFILER_EVENT_SCOPE(TEXT("FRenderStreamSyncFrameData::QuitNow()"));
+
     RenderStreamLink::instance().rs_setNewStatusMessage("");
+
+    IDisplayClusterClusterManager* ClusterMgr = IDisplayCluster::IsAvailable() ? IDisplayCluster::Get().GetClusterMgr() : nullptr;
+
+    if (ClusterMgr && !ClusterMgr->IsPrimary())
+    {
+        UE_LOG(LogRenderStream, Log, TEXT("RenderStream requested quit, waiting for nDisplay primary to terminate this node"));
```
josharms-disguise, line 246: Which versions was this happening on? Just 5.3? I believe I have fixed this before in a different way: https://github.com/disguise-one/RenderStream-UE/pull/132. This only went into 5.5 and above though.

josharms-disguise: Nevermind they seem different

josharms-disguise (approved):
LGTM but we should probably target RS3.0-UE5.5 since we weren't planning on making 3.0 releases for earlier versions.

pkevin-disguise (approved):
lgtm

Source/RenderStream/Private/SyncFrameData.cpp
josh-disguise: @elaineqiu-cmd was seeing it intermittently across 5.3 - 5.7. The only thing I don't know is why it was happening so much more on the branch we were testing, which only had a change to process spawning in it.

ivaylo-disguise (approved):

josh-disguise:
Retargetted to UE5.5

elaineqiu-cmd (commented):
QA passed.

PR #153, opened 2026-07-16
Refactor Maps Mode

josharms-disguise:
See [RSP-409](https://d3technologies.atlassian.net/browse/RSP-409?atlOrigin=eyJpIjoiYTUzZGEyMzY1ODQ2NDM0NWE0Yjg0MzBlMzU1NGY4YmEiLCJwIjoiaiJ9)

The Sphere requested a way to only save a certain selection of maps to the schema since the current behaviour is to save all the ones in cache which is every map the plugin can find even if it's not being used. UE already provides a list in the settings for "List of maps to include in packaged build." 

image

This feature just makes it so if that list is populated then we only save those ones to the schema, if it is not populated then it just results to the default behaviour.

[RSP-409]: https://d3technologies.atlassian.net/browse/RSP-409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
+++ b/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
@@ -40,6 +40,7 @@
 
 #include "FileHelpers.h"
 #include "GameMapsSettings.h"
+#include "Settings/ProjectPackagingSettings.h"
 
 #include "Logging/MessageLog.h"
 #include "Misc/UObjectToken.h"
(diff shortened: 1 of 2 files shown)


pkevin-disguise (approved):
nice, lgtm

ivaylo-disguise (approved):
lgtm

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -907,11 +908,33 @@ void FRenderStreamEditorModule::GenerateAssetMetadata()
...
+        if (bUseExplicitMapList)
+        {
+            for (const FFilePath& Map : PackagingSettings->MapsToCook)
+                PackagedMaps.Add(Map.FilePath);
+        }
+
+        TArray<const URenderStreamChannelCacheAsset*> MapsToSave;
+        for (const URenderStreamChannelCacheAsset* Cache : ChannelCaches)
+        {
+            if (!bUseExplicitMapList || PackagedMaps.Contains(Cache->Level.GetLongPackageName()))
```
anthonycheung-disguise, line 926: Could be a little more efficient if you did:
```
if(bUseExplicitMapList){
  [for loop]
}else{
  [set MapsToSave to reference ChannelCaches]
}
```

anthonycheung-disguise: lmk if I'm being too nitpicky tho :)

pkevin-disguise: The optimization wouldn't really have a measurable impact. We'd only save a couple of pointer copies at schema generation which is far from being a hot path.

PR #155, opened 2026-07-17
Fix UE failing to generate schema on "Save All"

egorovaanastasiia:
[RSP-411](https://d3technologies.atlassian.net/browse/RSP-411)

**Description:**
After [RSP-397](https://github.com/disguise-one/RenderStream-UE/pull/142) switched schema regeneration from save delegates to wrapping the `Save`/`SaveAllLevels` commands, the schema no longer regenerated for saves that don't go through those specific commands. Most visibly, using `Save All` after moving an object in a scene did nothing: the edit writes external-actor packages (not the level package), and that save path isn't one of the wrapped commands, so the schema went stale.

**Solution:**
- Restore the `PostSaveExternalActors` and `PostSaveWorldWithContext` editor delegates. These regenerate the schema whenever a save actually writes a world or external-actor package (Save All, content-browser saves, autosave, and OFPA actor edits). They complement the existing `Save`/`SaveAllLevels` command overrides, which still force a regenerate on an explicit save that writes nothing.
- Claude found another place for improvement: adding `RestoreSaveCommandOverrides()` call from `ShutdownModule`, to re-map the original command actions on module unload — so the global level-editor command list doesn't outlive the module holding a lambda that captures a freed module pointer.

**Note:**
Clicking `Save All` in UE only regenerates the schema after an object has been moved (different from `Save Current`). However, UE5.5 has exact same behaviour. RSP-397 was intended to match the behaviour between the 2.

[RSP-411]: https://d3technologies.atlassian.net/browse/RSP-411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSP-397]: https://d3technologies.atlassian.net/browse/RSP-397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
+++ b/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
@@ -94,6 +94,8 @@ void FRenderStreamEditorModule::StartupModule()
             this, &FRenderStreamEditorModule::RegisterSaveCommandOverrides));
     }
 
+    FEditorDelegates::PostSaveExternalActors.AddRaw(this, &FRenderStreamEditorModule::OnPostSaveWorld);
+    FEditorDelegates::PostSaveWorldWithContext.AddRaw(this, &FRenderStreamEditorModule::OnPostSaveWorldContext);
     FEditorDelegates::OnAssetsDeleted.AddRaw(this, &FRenderStreamEditorModule::OnAssetsDeleted);
     FCoreDelegates::OnBeginFrame.AddRaw(this, &FRenderStreamEditorModule::OnBeginFrame);
     FCoreDelegates::OnPostEngineInit.AddRaw(this, &FRenderStreamEditorModule::OnPostEngineInit);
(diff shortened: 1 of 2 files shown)


josharms-disguise (approved):
LGTM

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -1041,6 +1050,33 @@ void FRenderStreamEditorModule::RegisterSaveCommandOverrides()
     WrapSaveCommand(FLevelEditorCommands::Get().SaveAllLevels);
 }
 
+void FRenderStreamEditorModule::RestoreSaveCommandOverrides()
```
ivaylo-disguise, line 1053: It's not exactly clear to me if you found a use case where this restore is needed, or if you added it as defensive programming. I was under the impression that a situation where the lambda outlives the module is not possible.

ivaylo-disguise (approved):
looks good

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
egorovaanastasiia: No, didn't find a use case for this. Claude found it, I thought we should better have it there, just in case

elaineqiu-cmd (approved):
QA passed.

PR #156, opened 2026-07-31
Streaming Levels Refactor

josharms-disguise:
See [RSP-410](https://d3technologies.atlassian.net/browse/RSP-410?atlOrigin=eyJpIjoiMzI4OTU5YmZjYTcyNDBjZGEwNGEwZGZiYjdiZjM5OWYiLCJwIjoiaiJ9)

When working with many sublevels it can take awhile to save the schema since every exposed parameter in the persistent level is duplicated to every sublevel. This can cause friction when working in large projects and is an issue highlighted by the Sphere. The plugin side fixes are simple as we simply just don't save the persistent parameters to sublevel sections. A new schema field has been added so d3 can tell if the workload is using streaming levels or not. This also points to the scene index that it needs to pull from (for now it's always 0).

For the d3 changes see this [PR](https://github.com/disguise-one/d3/pull/7243)

[RSP-410]: https://d3technologies.atlassian.net/browse/RSP-410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
+++ b/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
@@ -472,8 +472,7 @@ void FetchLevelCaches(
 void GenerateScene(
     TMap<FSoftObjectPath, URenderStreamChannelCacheAsset*> const& LevelParams,
     RenderStreamLink::RemoteParameters& SceneParameters,
-    const URenderStreamChannelCacheAsset* Cache,
-    const URenderStreamChannelCacheAsset* Persistent)
+    const URenderStreamChannelCacheAsset* Cache)
 {
     FString sceneName = Cache->GetName();
     SceneParameters.name = _strdup(TCHAR_TO_UTF8(*sceneName));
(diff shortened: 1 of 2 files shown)


Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -889,6 +886,9 @@ void FRenderStreamEditorModule::GenerateAssetMetadata()
                 if (Cache != nullptr)
                     GenerateScene(LevelParams, *SceneParameters++, *Cache, MainMap);
             }
+
+            // Scene 0 is the persistent level
+            Schema.schema.baseSceneIndex = 0;
```
pkevin-disguise: is this something that can be changed in UE?

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
pkevin-disguise: Persistent is no longer used, could be removed. Also, the LevelParents in RenderstreamEditorModule.cpp:903-912 is used to build this parameter so it could be removed as well.

pkevin-disguise (changes requested):
- the min_d3_version might need to change since we change the schema

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
josharms-disguise: No, the persistent level always gets assigned 0 because of the order

ivaylo-disguise (approved):
lgtm

pkevin-disguise (approved):
lgtm

egorovaanastasiia (approved):

elaineqiu-cmd (approved):
QA passed.

PR #157, opened 2026-08-11
Packaging UE Builds With Shipping Config

josharms-disguise:
See [RSP-412](https://d3technologies.atlassian.net/browse/RSP-412?atlOrigin=eyJpIjoiYzk5YTI4ZmZkZGJiNDI4MTk1NTVkOTM4NDZmYzlkMjgiLCJwIjoiaiJ9)

There were 2 issues with packaging that the Sphere wanted resolved. One issue is that when packaging, by default UAT will produce 2 exes with the same name. One of these is the actually exe and one is a launcher known as the bootstrap exe. The solution is to use the -NoBootstrapExe flag to prevent this.

The second issue is that builds were being defaulted to the Development config. This is also a straightforward change but it produces consequences that need to be fixed as well. The first of these is that the resulting name that is given is appended with -win64-shipping. This is a problem since d3 expects the name of the exe to be the same as the json schema file name. The solution here is to just rename the exe after packaging to strip the end. The second consequence is that when in shipping mode UE blocks the ability to overwrite DefaultEngine.ini. This is a problem because on launch d3 will try to overwrite that file will specific entries needed for nDisplay to work correctly. The solution is to overwrite those files before the packaging is done.

[RSP-412]: https://d3technologies.atlassian.net/browse/RSP-412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Packaging now produces Shipping builds with prerequisite and bootstrap settings.
  * Missing engine and system configuration is added automatically.
  * Packaged executables are renamed to the project’s standard name, with metadata copied alongside them.

* **Bug Fixes**
  * Packaging post-processing now proceeds only after a successful build.
  * Existing target executables are replaced, and failures during renaming are reported.

### DIFF
--- a/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
+++ b/Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
@@ -948,8 +948,33 @@ struct FRequiredEngineSetting
     const TCHAR* Value;
 };
 
+bool IniSectionContainsSetting(const FString& IniContents, const FRequiredEngineSetting& Setting)
+{
+    const FString sectionHeader = FString::Printf(TEXT("[%s]"), Setting.Section);
+    const FString entry = FString::Printf(TEXT("%s=%s"), Setting.Key, Setting.Value);
+
+    TArray<FString> lines;
+    IniContents.ParseIntoArrayLines(lines);
+
+    bool inSection = false;
+    for (const FString& line : lines)
+    {
+        const FString trimmed = line.TrimStartAndEnd();
+
+        if (trimmed.StartsWith(TEXT(";")) || trimmed.StartsWith(TEXT("#")))
+            continue;
+
+        if (trimmed.StartsWith(TEXT("[")))
+            inSection = trimmed.Equals(sectionHeader, ESearchCase::IgnoreCase);
+        else if (inSection && trimmed.Equals(entry, ESearchCase::IgnoreCase))
+            return true;
+    }
+
+    return false;
+}
+
 // When laucnhing d3 tries to modifiy the ini files, however in shipping builds ini files are unable to be overwritten
-void EnsureShippingLaunchConfig()
+bool EnsureShippingLaunchConfig()
 {
     static const FRequiredEngineSetting RequiredSettings[] = {
         { TEXT("/Script/Engine.Engine"), TEXT("GameEngine"), TEXT("/Script/DisplayCluster.DisplayClusterGameEngine") },
@@ -967,7 +992,7 @@ void EnsureShippingLaunchConfig()
     for (const FRequiredEngineSetting& setting : RequiredSettings)
     {
         const FString entry = FString::Printf(TEXT("%s=%s"), setting.Key, setting.Value);
-        if (iniContents.Contains(entry))
+        if (IniSectionContainsSetting(iniContents, setting))
             continue;
 
         if (currentSection != setting.Section)
@@ -981,7 +1006,7 @@ void EnsureShippingLaunchConfig()
     }
 
     if (additions.IsEmpty())
-        return;
+        return true;
 
     // In case ini is already checked in
     if (SourceControlHelpers::IsEnabled() && FPaths::FileExists(engineIniPath))
@@ -992,7 +1017,12 @@ void EnsureShippingLaunchConfig()
     }
 
     if (!FFileHelper::SaveStringToFile(additions, *engineIniPath, FFileHelper::EEncodingOptions::ForceUTF8WithoutBOM, &IFileManager::Get(), FILEWRITE_Append))
+    {
         UE_LOG(LogRenderStreamEditor, Error, TEXT("Failed to write %s"), *engineIniPath);
+        return false;
+    }
+
+    return true;
 }
 
 void FRenderStreamEditorModule::RunPackageAndCopy()
@@ -1007,7 +1037,11 @@ void FRenderStreamEditorModule::RunPackageAndCopy()
     if(outputFolder == FString())
         return;
 
-    EnsureShippingLaunchConfig();
+    if (!EnsureShippingLaunchConfig())
+    {
+        UE_LOG(LogRenderStreamEditor, Error, TEXT("Aborting packaging, the required launch settings could not be written."));
+        return;
+    }
 
     FString arguments = FString::Printf(TEXT("Turnkey -command=VerifySdk -platform=Win64 -UpdateIfNeeded \
         BuildCookRun -nop4 -utf8output -nocompileeditor -skipbuildeditor -cook -project=\"%s\" -target=%s -unrealexe=\"%s\" \
@@ -1024,7 +1058,7 @@ void FRenderStreamEditorModule::RunPackageAndCopy()
     bool bSuccess = FPlatformProcess::ExecProcess(*uatPath, *arguments, &ReturnCode, &errorOut, nullptr);
     UE_LOG(LogTemp, Log, TEXT("Packaging complete..."));
 
-    if (bSuccess)
+    if (bSuccess && ReturnCode == 0)
     {
         // Need to copy metadata over to new .exe location
         FString filename = FString::Printf(TEXT("rs_%s.json"), FApp::GetProjectName());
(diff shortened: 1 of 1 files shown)


Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -940,6 +941,52 @@ FString FRenderStreamEditorModule::GetSelectedOutputFolder()
...
+            additions += FString::Printf(TEXT("%s[%s]%s"), LINE_TERMINATOR, setting.Section, LINE_TERMINATOR);
+        }
+
+        additions += entry + LINE_TERMINATOR;
+        UE_LOG(LogRenderStreamEditor, Log, TEXT("Adding [%s] %s to %s"), setting.Section, *entry, *engineIniPath);
+    }
+
+    if (additions.IsEmpty())
+        return;
+
```
ivaylo-disguise, line 1020: The .ini file may not be writeable under Perforce version control. You may have to checkout the file before writing it.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -952,10 +999,12 @@ void FRenderStreamEditorModule::RunPackageAndCopy()
     if(outputFolder == FString())
         return;
 
+    EnsureShippingLaunchConfig();
+
     FString arguments = FString::Printf(TEXT("Turnkey -command=VerifySdk -platform=Win64 -UpdateIfNeeded \
         BuildCookRun -nop4 -utf8output -nocompileeditor -skipbuildeditor -cook -project=\"%s\" -target=%s -unrealexe=\"%s\" \
         -platform=Win64 -installed -stage -archive -package -build -pak -iostore -compressed -prereqs \
```
ivaylo-disguise, line 1051: Claude flagged something interesting: "Shipping compiles out UE_LOG unless the target sets bUseLoggingInShipping = true"
Might be worth checking if we have fuctioning RenderStream logging when packaged for Shipping.

josharms-disguise: RenderStream logging also uses UE_LOG so unfortunately (or fortunately?) we don't get any. If we still wanted logs we would need some extra system to write out to a file I imagine.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -940,6 +941,52 @@ FString FRenderStreamEditorModule::GetSelectedOutputFolder()
...
+
+        additions += entry + LINE_TERMINATOR;
+        UE_LOG(LogRenderStreamEditor, Log, TEXT("Adding [%s] %s to %s"), setting.Section, *entry, *engineIniPath);
+    }
+
+    if (additions.IsEmpty())
+        return;
+
+    if (!FFileHelper::SaveStringToFile(additions, *engineIniPath, FFileHelper::EEncodingOptions::ForceUTF8WithoutBOM, &IFileManager::Get(), FILEWRITE_Append))
+        UE_LOG(LogRenderStreamEditor, Error, TEXT("Failed to write %s"), *engineIniPath);
```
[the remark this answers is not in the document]
pkevin-disguise: Packaging continues after SaveStringToFile fails, producing a Shipping build that cannot repair its own config at launch.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -940,6 +941,60 @@ FString FRenderStreamEditorModule::GetSelectedOutputFolder()
...
+
+    FString iniContents;
+    FFileHelper::LoadFileToString(iniContents, *engineIniPath);
+
+    FString additions;
+    FString currentSection;
+    for (const FRequiredEngineSetting& setting : RequiredSettings)
+    {
+        const FString entry = FString::Printf(TEXT("%s=%s"), setting.Key, setting.Value);
+        if (iniContents.Contains(entry))
```
pkevin-disguise: Existence check searches the whole file, so a commented-out or wrong-section entry could suppress a required setting.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -987,6 +1044,27 @@ void FRenderStreamEditorModule::RunPackageAndCopy()
         {
             UE_LOG(LogTemp, Log, TEXT("Failed to copy the meatadata over!"));
         }
+
+        // UAT adds a suffux to shipping builds that needs to be removed to match the json name
+        const FString binariesDir = outputFolder / FString::Printf(TEXT("Windows/%s/Binaries/Win64"), *projectName);
+        const FString undecoratedExe = binariesDir / FString::Printf(TEXT("%s.exe"), *projectName);
+
+        TArray<FString> decoratedExes;
+        IFileManager::Get().FindFiles(decoratedExes, *(binariesDir / TEXT("*-Win64-Shipping.exe")), true, false);
```
pkevin-disguise: *-Win64-Shipping.exe is quite widely reaching. I wonder if there is a way to make it a bit safer so that we don't match a random executable that matching this rule.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -987,6 +1036,27 @@ void FRenderStreamEditorModule::RunPackageAndCopy()
...
+
+        if (decoratedExes.Num() > 0)
+        {
+            const FString decoratedExe = binariesDir / decoratedExes[0];
+
+            if (FPaths::FileExists(undecoratedExe))
+                IFileManager::Get().Delete(*undecoratedExe);
+
+            if (!IFileManager::Get().Move(*undecoratedExe, *decoratedExe))
+                UE_LOG(LogRenderStreamEditor, Error, TEXT("Failed to rename %s to %s"), *decoratedExe, *undecoratedExe);
```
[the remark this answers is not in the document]
pkevin-disguise: I think we could log when we delete. Even better if we could log success/failure cases.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -1049,26 +1083,25 @@ void FRenderStreamEditorModule::RunPackageAndCopy()
...
         {
-            const FString decoratedExe = binariesDir / decoratedExes[0];
-
             if (FPaths::FileExists(undecoratedExe))
-                IFileManager::Get().Delete(*undecoratedExe);
+                UE_LOG(LogRenderStreamEditor, Log, TEXT("Replacing existing %s"), *undecoratedExe);
 
-            if (!IFileManager::Get().Move(*undecoratedExe, *decoratedExe))
+            if (IFileManager::Get().Move(*undecoratedExe, *decoratedExe))
+                UE_LOG(LogRenderStreamEditor, Log, TEXT("Renamed %s to %s"), *decoratedExe, *undecoratedExe);
```
pkevin-disguise, line 1097: if the rename fails, should we delete the old exe?

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -948,8 +948,33 @@ struct FRequiredEngineSetting
     const TCHAR* Value;
 };
 
+bool IniSectionContainsSetting(const FString& IniContents, const FRequiredEngineSetting& Setting)
```
pkevin-disguise: this is declared at file scope, should either be static or be in a namespace.

pkevin-disguise (changes requested):

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
josharms-disguise: Ya we probably should

pkevin-disguise (approved):
lgtm

elaineqiu-cmd (approved):
QA passed.

PR #158, opened 2026-08-18
Custom Icon For Packaging Button

josharms-disguise:
See [RSP-414](https://d3technologies.atlassian.net/browse/RSP-414?atlOrigin=eyJpIjoiZDBhMmExNDQzYjdjNGNiNjk5ZjAzNWVkZmQ2ZmJkNGEiLCJwIjoiaiJ9)

The button that was added to package projects has always used one of UE's default icons. It would be nice to use a custom one so that it's more obvious that the button is disguise related. A new custom icon was made for us so this change is just to assign that svg file as the button icon instead.

[RSP-414]: https://d3technologies.atlassian.net/browse/RSP-414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Style**
  * Added a dedicated icon for the RenderStream packaging toolbar button.
  * Updated the editor toolbar styling for a more consistent visual appearance.


egorovaanastasiia (approved):
lgtm.

Do we still support UE5.5? I thought we moved to UE5.6 since the UE5.8 plugin release

josharms-disguise:
> lgtm.
> 
> Do we still support UE5.5? I thought we moved to UE5.6 since the UE5.8 plugin release

Technically we don't but the packaging fixes that I just made were prompted by the Sphere and they specifically wanted the changes to go into 5.5. I figured that they would also want this change too since it's somewhat related.

pkevin-disguise (approved):
lgtm

Resources/PackagingIcon.svg
pkevin-disguise, line 1: actually, the existing resources are using pascal case for the file name.

also... the existing resource is the old disguise logo... maybe we could update it at the same time so that we don't have the new one and the old one in the same release.

pkevin-disguise (changes requested):

pkevin-disguise (approved):
:shipit:

elaineqiu-cmd (approved):
QA passed.

PR #166, opened 2026-08-26
RSP-415 Exposed parameters after an unsupported-typed variable read the wrong schema slot

pkevin-disguise:
RSP-415

Exposed parameters that follow an unsupported-typed variable no longer read the wrong schema slot.

### Root cause

Three walks iterate `GetProperties` in the same order and must agree on how many schema slots each property occupies: `GenerateParameters` (editor) emits them, `ValidateParameters` counts them, and `ApplyParameters` indexes into them.

`ApplyParameters` was the odd one out — it advanced `iParam` once per *property* rather than once per *schema slot*. The two disagree in three separate ways:

- unsupported properties (`String`, `Name`, non-texture object refs, unknown structs, non-double arrays) occupy **no** slot but advanced the index by one;
- vectors and rotators occupy **three** slots, colours **four**, but each advanced the index by one;
- custom events occupy **one** slot each and advanced the index by **none**.

Once the index drifted, the float-array branch read `(*ppParams)[iParam]` from the wrong entry — picking up another parameter's `nElements`, or reading past the end of the scene's parameter array entirely. Because `*ppParams += iParam` also feeds the next actor's starting offset, the drift compounded across every actor in the level. Validation passes in all of this, which is why the failure was silent.

### Change

- Added `SchemaParameterCount`, a single helper giving the slot count for a property, mirroring what validation and schema generation already do.
- `ApplyParameters` advances `iParam` by that count, via `ON_SCOPE_EXIT` so the existing out-of-range `continue` paths cannot skip the advance. Custom events now advance it too.
- Bounds-guarded the float-array schema read, and the caller now passes the number of *remaining* parameters rather than the scene total, so that guard is meaningful.

### Testing

- Invalid types should be skipped and logged
- Parameters that follows an invalid parameter should still work
- Exposed array that do not specify size generates a warning in the logs
  - format: `Skipping float array property %s: array is empty - size the array in the editor before exposing it`

### DIFF
--- a/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
+++ b/Source/RenderStream/Private/RenderStreamSceneSelector.cpp
@@ -19,6 +19,7 @@
 
 #include "Engine/Level.h"
 #include "Engine/World.h"
+#include "Misc/ScopeExit.h"
 #include "UObject/TextProperty.h"
 
 #include "VulkanRHIPrivate.h"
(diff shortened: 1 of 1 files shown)


Source/RenderStream/Private/RenderStreamSceneSelector.cpp
```diff
@@ -226,6 +227,55 @@ static bool validateField(FString key_, FString undecoratedSuffix, RenderStreamL
...
+        const FSoftObjectPath PropKey = SoftObjectProperty->GetPropertyValue(SoftObjectAddress).ToSoftObjectPath();
+        TSoftObjectPtr<USkeleton> Skeleton(PropKey);
+        return (Skeleton.IsValid() || Skeleton.IsPending()) ? 1 : 0;
+    }
+
+    if (Property->IsA(FTextProperty::StaticClass()))
+        return 1;
+
+    if (const FArrayProperty* ArrayProperty = CastField<const FArrayProperty>(Property))
+        return CastField<const FDoubleProperty>(ArrayProperty->Inner) ? 1 : 0;
```
ivaylo-disguise: A size check for the array may be needed here: if array is size zero, return count 0.

pkevin-disguise: done, logged a warning and we'll get the parameter mismatch error as well.

ivaylo-disguise (approved):
lgtm

egorovaanastasiia (approved):
lgtm

PR #168, opened 2026-09-16
PIE Display Is Minimized After Packaging Project

josharms-disguise:
See [RSP-419](https://d3technologies.atlassian.net/browse/RSP-419?atlOrigin=eyJpIjoiMWE3MDM0ZGFkMDUxNGI1OTlhM2FhNjFkYzA0NzA3YWUiLCJwIjoiaiJ9)

With the change to make all packaged projects Shipping, a workaround was needed to ensure the DefaultEngine.ini settings were correct. Normally when d3 launches a workload it uses the command line -ini feature which allows UE workloads to use additional settings without actually writing to the ini file. This feature is intended to be for debug builds only and is not compiled into Shipping builds. The workaround for this in the plugin was that we would append the needed settings to the ini file before packaging so there was no need to use the -ini argument. The issue is that the ini file never gets restored and it causes issues for the Play In Editor display since the editor is now using the modified ini to display its images. The result is that the image in the PIE is confined to the top left corner.

The solution is to instead write the settings we need to a file that the packed project will read. This makes it less complicated since we don't have to touch the DefaultEngine.ini that the Editor uses at all. Instead we write to WindowsEngine.ini which is the platform specific version that gets saved in the config folder of the packaged project.

[RSP-419]: https://d3technologies.atlassian.net/browse/RSP-419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Bug Fixes**
  - Packaging temporarily applies the required Windows launch settings before packaging and restores the project configuration afterward.
  - Existing project configuration content is preserved, including source-controlled files.
  - Leftover Windows launch configuration from an interrupted package is removed when the editor starts.
  - Packaging no longer writes launch settings to the staged packaged output after completion.


Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -1099,6 +1041,9 @@ void FRenderStreamEditorModule::RunPackageAndCopy()
 
     if (bSuccess && ReturnCode == 0)
     {
+        if (!RenderStreamPackaging::WriteShippingConfig(outputFolder, projectName))
```
egorovaanastasiia, line 1044: Seems like after moving the `WriteShipppingConfig()` logic here we run some guards only after the packaging. Is this intended?

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -979,87 +979,35 @@ FString FRenderStreamEditorModule::GetSelectedOutputFolder()
...
+            TEXT("GameEngine=/Script/DisplayCluster.DisplayClusterGameEngine") LINE_TERMINATOR
+            TEXT("GameViewportClientClassName=/Script/RenderStream.RenderStreamViewportClient") LINE_TERMINATOR
+            LINE_TERMINATOR
+            TEXT("[SystemSettings]") LINE_TERMINATOR
+            TEXT("rhi.UseSubmissionThread=0") LINE_TERMINATOR;
+
+        // Need to make sure this path doesn't already exist in the project
+        // If it does it will get put into the pak and UE will ignore the one we write
+        const FString projectConfig = FPaths::ConvertRelativePathToFull(FPaths::ProjectConfigDir() / TEXT("Windows/WindowsEngine.ini"));
+        if (FPaths::FileExists(projectConfig))
```
ivaylo-disguise, line 997: Question about this one: why do we expect WindowsEngine.ini not to exits? Isn't it possible for another plugin to write something to it before packaging, and we end up either aborting the build or losing another plugin's config?

Sounds a bit hacky, but I think a safer option would be to backup, modify .ini, package, restore .ini. The only risk is crashing mid-run and never restoring the .ini. But, this would make our config work with other parties' configs.

ivaylo-disguise: Also, there's aparently this option for building a config binary, which may force us to write the .ini in advance anyway.

josharms-disguise: I suppose it is possible that some other plugin would write to the windows file. The reason I initially didn't go with the restore method is for the reason you highlighted. It seems like a tradeoff either way. I guess the question is how likely is it that they will modify that file. I'll give the restore method another go because I don't want to have compatibility issues pop up later.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -977,90 +979,99 @@ FString FRenderStreamEditorModule::GetSelectedOutputFolder()
     return FString();
 }
 
-namespace RenderStreamPackaging
+// When launching a workload d3 passes these settings as command-line config overrides
+// Shipping builds don't allow this so we need to write them to somewhere the packaged project will read
+static const TCHAR* const ShippingConfig =
```
ivaylo-disguise, line 984: Do we need to add a new line at the beginning of the shipping config so when we append it to a non-empty file it starts from a new line?

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
```diff
@@ -977,90 +979,99 @@ FString FRenderStreamEditorModule::GetSelectedOutputFolder()
...
+private:
+    FString m_path;
+    TArray<uint8> m_original;
+    bool m_existed = false;
+    bool m_valid = false;
+};
+
+// A package interrupted by a crash doesn't restore the settings 
+// An exact match means the file is ours and can be safely removed
+void FRenderStreamEditorModule::RemoveStaleShippingConfig()
```
ivaylo-disguise, line 1069: So, not sure about this one, but Claude says `RunUAT BuildCookRun` is going to launch the unreal editor and call this, which would result  in the config that we just wrote being deleted. Worth a quick test I suppose.

josharms-disguise: Oh ya good catch. The cook process is an editor process so technically it loads the editor module and does the stale config check.

Source/RenderStreamEditor/Private/RenderStreamEditorModule.cpp
josharms-disguise: I think we probably should.

ivaylo-disguise (approved):
lgtm

egorovaanastasiia (approved):
lgtm

